PR #4, opened 2021-05-27
New public transport schema

haphut:
WIP


lassemaatta (commented):
> Reviewed 1 of 1 files at r1, 2 of 2 files at r2, 2 of 2 files at r3, 2 of 2 files at r4, 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4)* status: 9 of 15 files reviewed, all discussions resolved (waiting on @jarkkoka and @jschummer)

___
*[hasura/migrations/1620000000300_extensions/up.sql, line 10 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md2_Sek421dNuebiAKb:-Md2_Sek421dNuebiAKc:b80srpm) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000300_extensions/up.sql#L10)):*
> ```TSQL
> CREATE EXTENSION pgcrypto WITH SCHEMA extensions;
> CREATE EXTENSION postgis WITH SCHEMA extensions;
> ```

fyi: jore4-jore3-importer also uses the `btree_gist` extension for some "do these timeranges overlap?" constrainsts.

And also the [temporal tables](https://github.com/nearform/temporal_tables/commit/6ba58c78d68d59b2d95cfc22223729309f0c4c72) thing, but that's just a function rather than a full extension.

datarttu (commented):
Brilliant work in my opinion, goes hand in hand with Transmodel as good as it can, and provides a good basis for further development and extending of the network model! I wrote some more or less detailed comments, but some of them can be a bit out of the scope of this PR; anyway I think it's better they are there for reading rather than hanging around in my mind only. This is one of my first code reviews ever and I haven't been able to attend the architecture & design pattern discussions during the spring, so I might have missed some important points.

A couple of general details, probably not that critical though:

- Shall singular or plural table and view identifiers be preferred in general in Jore4? Plural identifiers are used here, which I personally find fine - I find it natural to refer to `infrastructure_links` since it is a "set of several things" by default. Using plural in the table / view names makes those identifiers more distinct from field names that tend to be in singular, making the code better readable. On the other hand, I think it is common too to refer to db tables and views using the name of a single entity, e.g. `infrastructure_link`, and this would go in line with the Transmodel class names. Whatever the choice is, I hope it is agreed upon as a common practice by the dev team together. Naming things is always a topic in itself...
- `IF NOT EXISTS` is used quite consistently throughout all the `CREATE something` statements, but I wonder if they are necessary / bring more benefits than risks here. I'm thinking of a scenario where we accidentally write a new migration with `CREATE TABLE IF NOT EXISTS foo(a int, b int)` while an older migration already included `CREATE TABLE foo(a int)`; now the new migration would not raise any error, suggesting that we should rather write `DROP TABLE ...` & `CREATE TABLE ...` or alternatively `ALTER TABLE foo ADD COLUMN b int` in the migration file, but just skip creating the `foo` table and leave it as it is, without the new column `b`.
- Almost the same applies to `OR REPLACE` when creating views and functions - I think if we leave it out, then we are in a good way forced to be explicit about what is dropped+created or altered in migration files. However, this would still not prevent us from accidentally overloading functions: say, we originally had function `do_foo(a int)`, then we forget to drop it before stating `CREATE FUNCTION do_foo(a int, b int)` in a new migration, and now we have both `do_foo(int)` and `do_foo(int, int) functions.

Now, those things listed may well have a justification that I just don't know about - as said, I don't find them critical things but am happy to learn about these things at least.

Good work all in all!

jschummer (commented):
Nice work in my opinion too and thank you @datarttu for your valuable comments :-)

IIRC, we did at some point come across the plural vs singular forms -discussion and IIRC, there were more people speaking in favour of the singular form. (Because even if the table "contains many rows", the table depicts a relation and maybe shouldn't be seen too much as being a container for things.)

I agree about the `IF NOT EXISTS` and `OR REPLACE`, at least if we really are to define the migrations' "down"-counterparts here too. And especially since we have the "Temporary destructive cleanup" for the rapid development phase, those `IF NOT EXISTS` should not be needed, I reckon.

When it comes to the different checks that could be added, I think they'd be good to have, even though I'm not sure how essential they are (due to a lack of understanding of Transmodel, I suppose). But it'd be good to implement the ones that can be implemented easily, nice findings there :-)

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-:-MfrQao95I9V2traYqfB:b-gx1z5a)* status: 9 of 15 files reviewed, 16 unresolved discussions (waiting on @datarttu, @haphut, and @jarkkoka)

___
*[hasura/migrations/1620000000400_utils/up.sql, line 21 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSp-r8-21:-Mfr-lPYEldIChK4NvcY:b1m3rm1) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000400_utils/up.sql#L21)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

Since `_ST_BestSRID` is an internal / helper function in PostGIS and not documented like normal PostGIS functions, I think a short comment on what this function does and where more information can be found would be appreciated here, if someone has to inspect this source code later. In fact, I tried to search for documentation of this function but did not find any good resources in a short time, so unfortunately I'm not able to provide a good link...

Generally, these wrappers seem to work fine, the code is well readable to a developer who knows the basics of PostGIS, and I find this implementation very elegant in general! Hopefully linear referencing of `geography` features will be supported officially in some future version of PostGIS, but until that we'll do well with these.
</blockquote></details>

I couldn't find any docs on `_ST_BestSRID` either.

On http://postgis.net/docs/PostGIS_Special_Functions_Index.html#PostGIS_TypeFunctionMatrix it is indicated that the functions marked with a smiley use `_ST_BestSRID`. However, if my browser's emoji support works correctly (...), there's currently not a single function making use of `_ST_BestSRID`.

Up until now, I didn't realize that `_ST_BestSRID` is an internal function (didn't know of the leading underscore convention). Now that I know, I'm not sure we should use it. On the other hand, AFAIR it is essential if we want to avoid hard-coding an SRID but still want to use the geometry type.

So I think you're right - if we decide to use it, we should at least document its use (here and maybe elsewhere) well. In case it breaks one day, we can probably still write our own Jore4-specific `_ST_BestSRID` by mapping the coordinates used by known instances use to specific SRIDs.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 46 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-46:-Mfqz4IIE91THKLjgWLZ:blbk39z) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L46)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

I wonder if this table is going to accommodate practical vehicle "classes" in the prod version of Jore4, such as buses of of types `A1`, `A2`, `C`, `D`, `MiniB`, etc., having different pax capacities, door configurations and so on. In such case I think it makes sense not to hard-code those values here. But for the first iteration of this table, these exemplary values here will do, I guess.

Or does this table have more general uses? Probably this will be connected somehow to the fleet registry / kalustorekisteri thingy which is of course an entire topic of its own then.
</blockquote></details>

If I understood correctly, the vehicle types here should only determine which infrastructure the vehicle can operate on. ('tall_electric_bus' e.g. cannot pass under certain bridges, even though it can otherwise use the same infra as 'generic_bus'.)

If we need distinction for other purposes (pax capacity, etc.) then that information should probably live elsewhere.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 72 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-72:-Mfrbde41cC2-zxAqjOg:b-zeok3n) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L72)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

I guess Transmodel does not discuss this and I'm not sure if this has been a topic among the developers, but I would like to see that creating duplicate links with equal (or equal as reversed) geometries would be restricted already at the database table level. I guess this could be done with a unique index on `shape` (and `ST_Reverse(shape)`) or with an `BEFORE INSERT OR UPDATE` trigger, taking into account a certain level of coordinate precision. Duplicate geometries are not easily seen in map UIs and can cause a lot of confusion before realizing what it is all about, in my experience (e.g. routes using different links that in fact represent the same street link shape). The random uuid primary key does not help here either since it just creates a new id for every new feature, whether it be imported from Jore3 or Digiroad or created by hand from a UI, without checking if the geometry and other attributes are actually distinct from the existing features.
</blockquote></details>

Some general comments related to the infrastructure part:

Since we decided we'll import the infrastructure from external sources, I'm not sure how much sense it makes to try to sanitize or comment on the data here.

If the externally provided data contains a duplicate link, should that prevent us from using it? Until now we only plan on correcting externally provided infra data by adding links to it (and marking existing links outdated), but removing part of the data is not planned yet.

In that sense, it might be better to make the model less restrictive and to have our import-backend make those checks you mention. Then we can let the user decide whether the found issues are affect the system in a way that it's better to not do the update at all.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 90 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-90:-MfrdPmPDJuBGGqR6-Qq:bgwt8zb) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L90)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

I'd find it reasonable to clarify here whether this is an estimation calculated from the shape geometry or created manually, e.g. based on field measurements.
</blockquote></details>

Same as above, I'm not sure we know for sure how these numbers got in there, since the data will be imported from an external source.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 250 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-250:-MfrjbSxDV1m_6XJXbGi:b-m677cv) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L250)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

Shouldn't `starts_from_scheduled_stop_point_id` and `ends_at_scheduled_stop_point_id` come from the first and last stops of journey patterns, or at least it should be enforced that these start and end points of a route cannot contradict with the first and last stops of journey patterns belonging to that route?
</blockquote></details>

I too am wondering if we can actually live without that duplicated information. :+1: 
___
*[hasura/migrations/1620000000700_authorize_roles/up.sql, line 20 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSv-r8-20:-MfrfrEZ2oGda-Kbw517:b-elcto9) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000700_authorize_roles/up.sql#L20)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

Using placeholder user names and replacing them with the [shell script](https://github.com/HSLdevcom/jore4-hasura/blob/main/scripts/replace-placeholders-in-sql-schema-migrations.sh) has probably been discussed in depth in the architecture meetings or otherwise; however, I would like to know what prevents us from defining group roles here, like `CREATE ROLE hasura_group NOLOGIN;`, granting them the privileges just like is done here, and then creating the actual users as members of the group roles when creating the services, e.g. `CREATE ROLE <hasura_username_secret_whatever> LOGIN PASSWORD '<hasura_db_password_secret>' IN ROLE hasura_group`. This way we would not need to manipulate the SQL source code strings.

The current implementation is elegant and functions obviously well at least as long as we don't have problems with string interpolation, or with illegal strings in secrets (as documented well in the root README). Anyway I'd like to learn more about this since I have not stumbled upon an implementation like this before.
</blockquote></details>

Using roles might be useful at some point, but I'm not sure how using roles would affect specifying secrets. After all the secrets would still need to be passed on from the cloud environment the same way. Maybe it's the "when creating the services" part which I don't understand :-)

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4)* status: 9 of 15 files reviewed, 17 unresolved discussions (waiting on @datarttu, @haphut, and @jarkkoka)

___
*[hasura/migrations/1620000000200_drop_public/up.sql, line 2 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Mge58-x9agzjSPC66ZG:-Mge58-x9agzjSPC66ZH:bxtlohp) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000200_drop_public/up.sql#L2)):*
> ```TSQL
> -- Clarify that public does not contain anything valuable.
> DROP SCHEMA IF EXISTS public CASCADE;
> ```

we cannot drop the public schema in Azure as it is not owned by the SQL admin user we have

jarkkoka (commented):
Indeed, very valuable commenrs from @datarttu . Thanks for your feedback and insight!

> Reviewed 1 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-:-Mjdl3tNEeSHPkufP0Gl:bmdbziv)* status: 8 of 15 files reviewed, 17 unresolved discussions (waiting on @datarttu, @haphut, and @lassemaatta)

___
*[hasura/migrations/1620000000400_utils/up.sql, line 21 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSp-r8-21:-MjdiV032keqwWO4F2fD:b-ult890) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000400_utils/up.sql#L21)):*
> IIRC

I think I get pretty weird return value (`999035`) from `_ST_BestSRID` function while giving it point geographies that reside in Helsinki city region as parameter. With PostGIS extension v3.1.3 (using PostgreSQL v13.4), the `spatial_ref_sys`table does not seem to contain entry with SRID value `999035`. Since the private function is not documented, I am not sure whether the return value should be strictly interpreted as SRID value or something else.

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4)* status: 8 of 15 files reviewed, 17 unresolved discussions (waiting on @datarttu, @haphut, and @lassemaatta)

___
*[hasura/migrations/1620000000400_utils/up.sql, line 21 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSp-r8-21:-MjdrQzW6W_SZN5jPrPE:bsg74bn) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000400_utils/up.sql#L21)):*

> IIRC

I think I get pretty weird return value (`999035`) from `_ST_BestSRID` function while giving it point geographies that reside in Helsinki city region as parameter. With PostGIS extension v3.1.3 (using PostgreSQL v13.4), the `spatial_ref_sys`table does not seem to contain entry with SRID value `999035`. Since the private function is not documented, I am not sure whether the return value should be strictly interpreted as SRID value or something else.
</blockquote></details>

I found a ten year old PostGIS ticket (https://trac.osgeo.org/postgis/ticket/1230) where the (likely non-widely-recognized) `999` series of SRID values is kind of documented in one comment as follows:

 ```
999001 to 999060 for UTM North (zone: 1 to 60) 
```

So `999035` would translate to UTM zone 35 (north) which I think corresponds to more known `EPSG:3067` that is generally used in Finland. So I think it seems ok, if the meaning has remained unchanged. Maybe we should find more documentation to be on a safe ground.

**Review by pallost — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4)* status: 8 of 15 files reviewed, 18 unresolved discussions (waiting on @datarttu, @haphut, and @lassemaatta)

___
*[hasura/migrations/1620000000700_authorize_roles/up.sql, line 57 at r9](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MjmjqHk8yMTdHVcykYO:-MjmjqHk8yMTdHVcykYP:b-2cl3li) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/bddb4387a83754ee8dc7e2d971c164d023cfac54/hasura/migrations/1620000000700_authorize_roles/up.sql#L57)):*
> ```TSQL
>   journey_pattern
>   GRANT USAGE ON SEQUENCES
>   TO xxx_hsl_jore4_db_hasura_username_xxx;
> ```

should add the requirement to the Docker reference in the README that we need the hsl-jore4-db-hasura-username secret to be set for the container

**Review by ossiberg — COMMENTED** [review]
> Reviewed 2 of 2 files at r5, 1 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4)* status: 10 of 15 files reviewed, 18 unresolved discussions (waiting on @datarttu, @haphut, @jarkkoka, @jschummer, @lassemaatta, @ossiberg, and @pallost)

___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 54 at r9](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MkLl8CF6i0AmCh-pcdr:-MkLl8CF6i0AmCh-pcds:bkjapgh) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/bddb4387a83754ee8dc7e2d971c164d023cfac54/hasura/migrations/1620000000500_initial_public_transport/up.sql#L54)):*
> ```TSQL
>   ('generic_metro', 'metro'),
>   ('generic_ferry', 'ferry'),
>   ('tall_electric_bus', 'bus')
> ```

FYI, I suppose this is on line/route level. A single line could be operated with multiple vehicle types, such as electric bus and diesel bus. Therefore I don't find it convenient to specify tall_electric_bus here

**Review by pallost — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4)* status: 4 of 17 files reviewed, 19 unresolved discussions (waiting on @datarttu, @haphut, @jarkkoka, @jschummer, @lassemaatta, @ossiberg, and @pallost)

___
*[hasura/migrations/1620000000300_extensions/up.sql, line 26 at r13](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MkvFoAg4QZnXtAMxPi-:-MkvFoAg4QZnXtAMxPi0:bfv9gex) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/b48426177bcc3ea035b16bd9be62d102378d9b1a/hasura/migrations/1620000000300_extensions/up.sql#L26)):*
> ```TSQL
> DO $$ BEGIN
>   EXECUTE format(
>     'ALTER DATABASE %I SET search_path = extensions',
> ```

pgcrypto has to be in the search_path BEFORE the hasura migrations are executed: https://hasura.io/docs/latest/graphql/core/deployment/postgres-requirements.html#pgcrypto-in-pg-search-path So the order goes such:

1. something/someone sets up pgcrypto extension to some schema (e.g. jore4-deploy could set it up...?). If not, hasura will set it up for itself to the public schema
2. hasura runs its own migrations to create its own system (which uses at least uuid types from pgcrypto)
3. hasura runs the migrations from the /hasura/migrations folder

Personally I see no issues with just having the extensions in the public schema... but if you need them to be somewhere else, maybe they should be added to this playbook: https://github.com/HSLdevcom/jore4-deploy/blob/main/ansible/play-configure-database.yml#L28
e.g. by using https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_schema_module.html#ansible-collections-community-postgresql-postgresql-schema-module and https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_ext_module.html#ansible-collections-community-postgresql-postgresql-ext-module

**Review by pallost — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4)* status: 4 of 17 files reviewed, 20 unresolved discussions (waiting on @datarttu, @haphut, @jarkkoka, @jschummer, @lassemaatta, @ossiberg, and @pallost)

___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 88 at r13](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MkvIUOi17vV54EJTP8h:-MkvIUOi17vV54EJTP8i:b-4o39tl) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/b48426177bcc3ea035b16bd9be62d102378d9b1a/hasura/migrations/1620000000500_initial_public_transport/up.sql#L88)):*
> ```TSQL
>   'The infrastructure links, e.g. road or rail elements: https://www.transmodel-cen.eu/model/index.htm?goto=2:1:1:1:453';
> COMMENT ON COLUMN
>   infrastructure_network.infrastructure_link.infrastructure_link_id IS
> ```

is this the jore4 id for the link or the digiroad id? Do we need to store the external id (=digiroad id) as well for reference?

**Review by jschummer — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4)* status: 4 of 18 files reviewed, 20 unresolved discussions (waiting on @datarttu, @jarkkoka, @jschummer, @lassemaatta, @ossiberg, and @pallost)

___
*[hasura/migrations/1620000000300_extensions/up.sql, line 10 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md2_Sek421dNuebiAKb:-MkvyY0aBAZE9IRcFV7Z:bwgqqth) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000300_extensions/up.sql#L10)):*
<details><summary><i>Previously, lassemaatta (Lasse Määttä) wrote…</i></summary><blockquote>

fyi: jore4-jore3-importer also uses the `btree_gist` extension for some "do these timeranges overlap?" constrainsts.

And also the [temporal tables](https://github.com/nearform/temporal_tables/commit/6ba58c78d68d59b2d95cfc22223729309f0c4c72) thing, but that's just a function rather than a full extension.
</blockquote></details>

These should probably be taken care of in their own migrations, when they are needed.
___
*[hasura/migrations/1620000000300_extensions/up.sql, line 26 at r13](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MkvFoAg4QZnXtAMxPi-:-MkvygGW3Zyp0pSdZrLF:b1m5fov) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/b48426177bcc3ea035b16bd9be62d102378d9b1a/hasura/migrations/1620000000300_extensions/up.sql#L26)):*

pgcrypto has to be in the search_path BEFORE the hasura migrations are executed: https://hasura.io/docs/latest/graphql/core/deployment/postgres-requirements.html#pgcrypto-in-pg-search-path So the order goes such:

1. something/someone sets up pgcrypto extension to some schema (e.g. jore4-deploy could set it up...?). If not, hasura will set it up for itself to the public schema
2. hasura runs its own migrations to create its own system (which uses at least uuid types from pgcrypto)
3. hasura runs the migrations from the /hasura/migrations folder

Personally I see no issues with just having the extensions in the public schema... but if you need them to be somewhere else, maybe they should be added to this playbook: https://github.com/HSLdevcom/jore4-deploy/blob/main/ansible/play-configure-database.yml#L28
e.g. by using https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_schema_module.html#ansible-collections-community-postgresql-postgresql-schema-module and https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_ext_module.html#ansible-collections-community-postgresql-postgresql-ext-module
</blockquote></details>

Hasura actually creates pgcrypto in the public schema if it doesn't find it at startup.

If it works out, I think the separate extensions-schema is quite neat, so I now added lines to drop the used extensions (see above).
___
*[hasura/migrations/1620000000400_utils/up.sql, line 21 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSp-r8-21:-Mkvz69O6baCHungwHit:b-w6xvag) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000400_utils/up.sql#L21)):*

I found a ten year old PostGIS ticket (https://trac.osgeo.org/postgis/ticket/1230) where the (likely non-widely-recognized) `999` series of SRID values is kind of documented in one comment as follows:

 ```
999001 to 999060 for UTM North (zone: 1 to 60) 
```

So `999035` would translate to UTM zone 35 (north) which I think corresponds to more known `EPSG:3067` that is generally used in Finland. So I think it seems ok, if the meaning has remained unchanged. Maybe we should find more documentation to be on a safe ground.
</blockquote></details>

I now added wrapper functions for _ST_BestSRID (determine_SRID), so we have a single point of change (actually two functions..) in case _ST_BestSRID stops working one day. If nothing else, one can return a static value from those.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 46 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-46:-MkvzPtM5CHnPk5R5WvL:b-tl259) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L46)):*

If I understood correctly, the vehicle types here should only determine which infrastructure the vehicle can operate on. ('tall_electric_bus' e.g. cannot pass under certain bridges, even though it can otherwise use the same infra as 'generic_bus'.)

If we need distinction for other purposes (pax capacity, etc.) then that information should probably live elsewhere.
</blockquote></details>

When discussing this with the team, we came to the conclusion that these definitions should indeed only specify some infra-related things about a vehicle. Thus they are actually not the "vehicle type" as mentioned in Transmodel, but the mapping was renamed to "vehicle_sub_mode".
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 73 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-73:-Mkw-FWjEegToZo2KMYh:b-sey9lw) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L73)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

Thinking of a topologically consistent, routable network, I would expect to see something like `start_node` and `end_node` here, referring to [infrastructure point](http://www.transmodel-cen.eu/model/index.htm?goto=2:1:1:1:449) ids and telling which links are connected to each other, which in turn enables creating valid route paths along the network and finding intersections connecting more than two links, for instance. Is such a feature planned to be implemented later?
</blockquote></details>

According to our current plan, our own infra will not contain the topology, but we'll only "actively" process links. The postgresql pgrouting extension will create the topology from those and will find routes on our links.

If we need to e.g. identify all links connecting to a crossroads, that should be doable with this setup (either through pgrouting or by selecting the links in an area around the crossroads).
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 74 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-74:-Mkw-ytI1_7hBGVtC5WZ:bglw02v) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L74)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

Generally, this "three-valued logic" with link directionality is a bit confusing in my opinion. Is there a clear reason to allow traversing oneway links either to the digitized linestring direction or to the opposite direction, or could we just define that oneway links can only be traversed according to the linestring direction? Of course the Digiroad data model supports both digitized and opposite direction with oneway links, which makes this not that straightforward... I wonder in general why the geometries of those opposite cases have simply not been reversed in Digiroad from the very beginning.

I see two scenarios:

1. Accept the current implementation that allows importing Digiroad links as they are, even the oneway links whose allowed traversal direction is marked as opposite to the digitized direction. This keeps the import a bit more simple and the link data in Jore4 remains more consistent with original Digiroad data. On the other hand, this behaviour should be documented very clearly to both developers who should understand what `is_direction_forwards` really means (`NULL` as two-way traffic allowed can be a bit unintuitive at first) and Jore4 users / network operators who must be aware that the geometry direction of oneway links is not always the same as the allowed traversal direction (-> risk of confusion). Also, these oneway links must be treated correctly with `ST_Reverse()` in views and functions / procedures using this table, as seen [below](https://github.com/HSLdevcom/jore4-hasura/pull/4/files#diff-7a91f43d516c08739a746198a0bf2842c5c9b4ef74d19294c3b33749461b599fR336). Probably not a big performance problem, but adds more code to maintain at least.
2. Define that oneway `infrastructure_links` shapes in Jore4 must always be created in the same direction as the allowed traversal direction. This requires that the "opposite direction allowed" oneway cases among Digiroad links get their `ST_Reverse()` treatment upon the Digiroad -> Jore4 import, and of course this would be something to be aware of when comparing Jore4 and Digiroad data models in general. Now the allowed traversal directions could be modeled with e.g. `is_twoway boolean DEFAULT true NOT NULL`, which could be more transparent to developers and leave less room for user (modifying network) errors. No `CASE WHEN` & `ST_Reverse()` would be needed for oneway links in views, functions and procedures.

Unfortunately Transmodel does not seem to instruct how to implement this directionality thing, at least I did not find anything.
</blockquote></details>

We discussed this and agreed. Thus this has been changed to an enum (but not a postgresql enum, since Hasura cannot handle those).
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 94 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-94:-Mkw0Hzx6dCJTjrJ1WnS:bnwp41h) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L94)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

I like informative identifiers but find this one already pretty long. I'd suggest naming the table something like `vehicle_types_on_infrastructure_links` and keeping the `COMMENT ON TABLE` here as it is now.
</blockquote></details>

Good suggestion, implemented it :-)
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 113 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-113:-Mkw0OHi1Bhn2vzD0VMI:b-oib351) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L113)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

Shall we (later) add a view like `infrastructure_network.infrastructure_links_safely_traversed_by_vehicle_modes` that connects `infrastructure_network.infrastructure_links_safely_traversed_by_vehicle_types` and `reusable_components.vehicle_types`, showing which links can be traversed by which vehicle _modes_ such as bus, tram, metro, train, ferry? Consequently, I would find it nice to allow setting vehicle types allowed on a link through the `belonging_to_vehicle_mode` parent rather than inserting vehicle types separately. This of course should be evaluated through the use cases - whether it makes sense to do this "vehicle types first" or "modes first", or both.
</blockquote></details>

We came to the conclusion that the vehicle_sub_mode can be used for this. It seems like the "vehicle type", which was used earlier, caused some misunderstandings. But yes, adding an additional view to sum up things may be a good option later if needed.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 144 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-144:-Mkw0oWH851hN2l7b8C3:bvahi3i) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L144)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

Could we use `INSTEAD OF INSERT / UPDATE / DELETE` triggers on the `service_pattern.scheduled_stop_points` view here, making the view visible as a table-like object that allows conventional DML statements to the server layer and other database users, rather than custom functions that the server / other user must call to insert, update or delete data? I understand if `INSTEAD OF` triggers are considered too non-standard / confusing design by developers, though. [Here](https://github.com/datarttu/sujuikoDB/blob/dev/db/020_schema_nw.sql) are some examples on using such triggers to manage network link and node data.
</blockquote></details>

This might be a good option, also because Hasura does not like to add mutations from custom functions. (Maybe this works better in v2, but I am not sure).

So in this sense, the instead-of-triggers may actually solve two different problems at once. However, I'd prefer to add those in a separate PR.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 170 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-170:-Mkw1OWe0Xxh0OsdpAqd:bpg4lmx) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L170)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

I wonder if this could be handled without CTE, like this:

```
SELECT
  ssp.scheduled_stop_point_id,
  ssp.label,
  ssp.measured_location,
  ssp.located_on_infrastructure_link_id,
  ssp.is_direction_forwards_on_infrastructure_link,
  il.shape,
  internal_utils.ST_LineLocatePoint(il.shape, ssp.measured_location) AS relative_distance_from_infrastructure_link_start,
  internal_utils.ST_ClosestPoint(il.shape, ssp.measured_location) AS closest_point_on_infrastructure_link
FROM
  internal_service_pattern.scheduled_stop_points AS ssp
  INNER JOIN infrastructure_network.infrastructure_links AS il ON (ssp.located_on_infrastructure_link_id = il.infrastructure_link_id);
```

This would of course require writing a `geography` wrapper of `ST_ClosestPoint` in the `internal_utils` schema, but it might well have other use cases than only this, too. `ST_LineInterpolatePoint` after the current CTE block seems a bit like round-trip to me, while we could get the same point geometry, based on the link shape and closest point, directly with `ST_ClosestPoint`.
</blockquote></details>

I've now changed that (and added our own ST_ClosestPoint), nice finding!
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 250 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-250:-Mkw1UaF-vfK-zVqK4Sy:b-gws2wh) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L250)):*

I too am wondering if we can actually live without that duplicated information. :+1: 
</blockquote></details>

After thinking this through, I actually see that there is no (not much) duplicate information. There maybe more than one journey pattern per route and I'm also not sure if a journey pattern could only include part of a route and therefore the start and end stop points might not match.

But I do think that this needs a constraint trigger to check whether the start and end stop points of a route are on the route's (first and last) links.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 271 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-271:-Mkw2GON60OhCw6ANyqk:b-2741xt) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L271)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

I'd suggest `CHECK (infrastructure_link_sequence > 0)` here, unless we really have use cases for negative sequence numbers.
</blockquote></details>

To be honest I don't see the value of this restriction. I also don't think that there's value in serving continuous sequence numbers (as suggested elsewhere). I don't see a problem as long as there are no duplicates (which is guaranteed by the index).
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 297 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-297:-Mkw3IP92_aoU5OH1AWu:b1bgvyu) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L297)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

I would use here a trigger that forces the sequence number to behave correctly within each `route_id` (1 ... N in steps of 1, where N is the total number of links along the route), so that the underlying data really represents the business need, not only a view on top of it. Whether this or the view approach would be easier to apply and maintain should be discussed in more detail I guess.
</blockquote></details>

I was first agreeing with that and actually implemented it in a view. But that I realised that editing a mid-segment of the route would become impossible, at least with the view approach.

Assume, there's links with seq numbers 1, 5, 7. The frontend sees those as 1, 2, 3, so it cannot (or does not know where to) insert the new data.

I think the trigger solution could work, but I'd rather leave that to a separate PR (more triggers to come and integration tests for them).
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 298 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-298:-Mkw32C369GOPfqoGTcm:bwrstll) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L298)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

I also find it important to ensure that link with seq N+1 starts from the same point where link with seq N ends, so that the route really forms a continous, gapless path and especially two-way links are not used with a wrong traversal direction by the route.
</blockquote></details>

That is something which we'd rather leave up to pgrouting. We'd rather not manage the topology ourselves.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 429 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-429:-Mkw4D1NA0qaZVo1hTci:b-39j0ml) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L429)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

And check here also (perhaps by the same constraint) that both `scheduled_stop_point_sequence` and `infrastructure_link_sequence` are increasing along the route, i.e. "stop `scheduled_stop_point_sequence = 5` cannot be visited by link `infrastructure_link_sequence = 50` if link `infrastructure_link_sequence = 48` already visited stop `scheduled_stop_point_sequence = 4`.
</blockquote></details>

Triggers for that are underway and will come in a separate PR :-) 
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 54 at r9](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MkLl8CF6i0AmCh-pcdr:-MkvzyxjAG1tpxxVmyxY:b-1z7u6d) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/bddb4387a83754ee8dc7e2d971c164d023cfac54/hasura/migrations/1620000000500_initial_public_transport/up.sql#L54)):*
<details><summary><i>Previously, ossiberg (Ossi Berg) wrote…</i></summary><blockquote>

FYI, I suppose this is on line/route level. A single line could be operated with multiple vehicle types, such as electric bus and diesel bus. Therefore I don't find it convenient to specify tall_electric_bus here
</blockquote></details>

Until now these are only used to specify which "vehicle_sub_mode" can traverse which links of the infra. We'll try to keep in mind that the line can be traversed by different vehicle_sub_modes :+1: 
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 88 at r13](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MkvIUOi17vV54EJTP8h:-Mkw07aeB2Lp_fF6B4w2:b-agqoj5) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/b48426177bcc3ea035b16bd9be62d102378d9b1a/hasura/migrations/1620000000500_initial_public_transport/up.sql#L88)):*

is this the jore4 id for the link or the digiroad id? Do we need to store the external id (=digiroad id) as well for reference?
</blockquote></details>

This is our own link id. The reference to an infra source (e.g. digiroad) should be added later.
___
*[hasura/migrations/1620000000700_authorize_roles/up.sql, line 57 at r9](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MjmjqHk8yMTdHVcykYO:-Mkw4NFZ27-uSCoo2c9n:bxemhpb) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/bddb4387a83754ee8dc7e2d971c164d023cfac54/hasura/migrations/1620000000700_authorize_roles/up.sql#L57)):*

should add the requirement to the Docker reference in the README that we need the hsl-jore4-db-hasura-username secret to be set for the container
</blockquote></details>

This was corrected to refer to the "db-username" secret. Also added references to db-auth-username and db-jore3importer-username, which were missing.
___
*[hasura/migrations/1620000000200_drop_public/up.sql, line 2 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Mge58-x9agzjSPC66ZG:-Mkw4J462KQJwwaStPn-:b-4nytz4) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000200_drop_public/up.sql#L2)):*

we cannot drop the public schema in Azure as it is not owned by the SQL admin user we have
</blockquote></details>

True, this was fixed (removed the drop)!

jarkkoka (commented):
Basically, looks pretty good to me. We may augment the data model as further requirements arise. 

> Reviewed 1 of 7 files at r10, 1 of 2 files at r16, 2 of 2 files at r17, 6 of 6 files at r18.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-:-MlAmddW0vNr_MBjtx7B:bt5lptr)* status: 14 of 18 files reviewed, 26 unresolved discussions (waiting on @datarttu, @haphut, @jarkkoka, @jschummer, @lassemaatta, @ossiberg, and @pallost)

___
*[hasura/migrations/1620000000400_utils/up.sql, line 23 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlAceBp8EREjJe_rzgR:-MlAceBp8EREjJe_rzgS:b-uuv00m) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000400_utils/up.sql#L23)):*
> ```
> Determine the SRID to be used for the given geography
> ```

"Determine the most suitable SRID to be used for the given geography when converting it to a geometry." or something similar a bit more descriptive.
___
*[hasura/migrations/1620000000400_utils/up.sql, line 43 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlAd4Qa5plsyLTo42gE:-MlAd4Qa5plsyLTo42gF:b-vyk2yk) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000400_utils/up.sql#L43)):*
> ```
> Determine the SRID to be used for the given geographies.
> ```

"Determine the most suitable common SRID to be used for the given geographies when converting them into geometries." or something similar a bit more descriptive.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 90 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-90:-MlAg-tj7tUr6ADFkGoQ:b-ioau5q) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L90)):*

Same as above, I'm not sure we know for sure how these numbers got in there, since the data will be imported from an external source.
</blockquote></details>

In case of Digiroad links, the length is calculated from the geometry object.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 33 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlAeHdt43vVlq6kqbbm:-MlAeHdt43vVlq6kqbbn:bnvh697) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000500_initial_public_transport/up.sql#L33)):*
> ```
> vehicle_sub_mode
> ```

I'd rather have a table named `vehicle_submode` (without underscore in between) since the word "sub" means kind of nothing in itself.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 39 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlAefel-GpWgRNliMXU:-MlAefel-GpWgRNliMXV:b5hd72y) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000500_initial_public_transport/up.sql#L39)):*
> ```
> sub mode
> ```

Replace "sub mode" with "submode".
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 280 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlAYrLI8JKPbnA-21d_:-MlAYrLI8JKPbnA-21da:bfwppde) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000500_initial_public_transport/up.sql#L280)):*
> ```
> infrastruture
> ```

Typo. Fix => "infrastructure"
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 325 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlA_7rMAgXhGRSCwan-:-MlA_7rMAgXhGRSCwan0:b-dww6az) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000500_initial_public_transport/up.sql#L325)):*
> ```
>     r.description_i18n,
>     r.starts_from_scheduled_stop_point_id,
>     r.ends_at_scheduled_stop_point_id,
> ```

Shouldn't these three non-aggregate columns also appear in GROUP BY clause, should they?

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4)* status: 14 of 18 files reviewed, 28 unresolved discussions (waiting on @datarttu, @haphut, @jarkkoka, @jschummer, @lassemaatta, @ossiberg, and @pallost)

___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 285 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlApfDfDdXOhWncJlaP:-MlApfDfDdXOhWncJlaQ:brqbbfv) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000500_initial_public_transport/up.sql#L285)):*
> ```
> CREATE INDEX
> ```

Maybe this should be `CREATE UNIQUE INDEX`?
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 417 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlAptVRASa1W8Z0WFSy:-MlAptVRASa1W8Z0WFSz:brqbbfv) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000500_initial_public_transport/up.sql#L417)):*
> ```
> CREATE INDEX
> ```

Maybe this should be `CREATE UNIQUE INDEX`?

jschummer (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4)* status: 14 of 18 files reviewed, 28 unresolved discussions (waiting on @datarttu, @jarkkoka, @jschummer, @lassemaatta, @ossiberg, and @pallost)

___
*[hasura/migrations/1620000000400_utils/up.sql, line 23 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlAceBp8EREjJe_rzgR:-MlHCViO2Do2fgD0KTOU:bef2680) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000400_utils/up.sql#L23)):*
> Determine the most suitable common SRID to be used for the given geographies when converting them into geometries.

Basically it doesn't need to be the most suitable SRID (according to any definition of suitable). Like mentioned in the end of the comment, in many scenarios it may be sufficient to return a constant for all cases.

Since at the moment, it does use _ST_BestSRID, however, I think it makes sense to change the comment to what you suggested and that's what I did :-)
___
*[hasura/migrations/1620000000400_utils/up.sql, line 43 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlAd4Qa5plsyLTo42gE:-MlHCzAS3UVGvG_KSJL9:bt0t1my) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000400_utils/up.sql#L43)):*

> ```
> Determine the SRID to be used for the given geographies.
> ```

"Determine the most suitable common SRID to be used for the given geographies when converting them into geometries." or something similar a bit more descriptive.
</blockquote></details>

Same goes as above.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 90 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-90:-MlHDnrA5GtOvbK7O8RV:b-jqpywc) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L90)):*

In case of Digiroad links, the length is calculated from the geometry object.
</blockquote></details>

Yes, but I'd still rather not mention that here, since digiroad is only one source of infra, even though maybe the most important one. So this info is maybe better picked up from the Digiroad specs, not ours.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 144 at r8](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-Md-HCfuSMk1jLHtmMSr-r8-144:-MlHE8aAAh3PBdDBiGxC:b-dgs3x2) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/21d9ec533a6bcd1db53a9f2b02427bcbf93ff805/hasura/migrations/1620000000500_initial_public_transport/up.sql#L144)):*

This might be a good option, also because Hasura does not like to add mutations from custom functions. (Maybe this works better in v2, but I am not sure).

So in this sense, the instead-of-triggers may actually solve two different problems at once. However, I'd prefer to add those in a separate PR.
</blockquote></details>

I actually did add the INSTEAD OF triggers now. The integration tests for those are still due, however.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 33 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlAeHdt43vVlq6kqbbm:-MlHD1qFBu5YFSfdG_t3:b-257vw8) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000500_initial_public_transport/up.sql#L33)):*

> ```
> vehicle_sub_mode
> ```

I'd rather have a table named `vehicle_submode` (without underscore in between) since the word "sub" means kind of nothing in itself.
</blockquote></details>

I somewhat disagree in the sense that in my opinion, also prefixes could be separated by underscores (or otherwise). Prominent examples might be "pre" and "post".

That being said, submode might still make more sense here as it's less likely to be confused with a metro ("sub" in some parts of the world). Also, we don't have a "super_mode" or similar. So I did change it :+1: 
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 39 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlAefel-GpWgRNliMXU:-MlHDl8XDv8MqV1nAlLf:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000500_initial_public_transport/up.sql#L39)):*

> ```
> sub mode
> ```

Replace "sub mode" with "submode".
</blockquote></details>

Done.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 280 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlAYrLI8JKPbnA-21d_:-MlHEHNYBufYBT9EbvvV:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000500_initial_public_transport/up.sql#L280)):*

> ```
> infrastruture
> ```

Typo. Fix => "infrastructure"
</blockquote></details>

Done.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 285 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlApfDfDdXOhWncJlaP:-MlHEJSf21bFGgAm3sBd:b-unnm7z) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000500_initial_public_transport/up.sql#L285)):*

> ```
> CREATE INDEX
> ```

Maybe this should be `CREATE UNIQUE INDEX`?
</blockquote></details>

I think it shouldn't. As the comment above states, the primary key includes a unique constraint for those columns, the other way round.

So this index is just to speed up look ups based on the opposite column order.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 325 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlA_7rMAgXhGRSCwan-:-MlHFuPu5ZKI_DDiL3ZJ:b-4fu7e0) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000500_initial_public_transport/up.sql#L325)):*

> ```
>     r.description_i18n,
>     r.starts_from_scheduled_stop_point_id,
>     r.ends_at_scheduled_stop_point_id,
> ```

Shouldn't these three non-aggregate columns also appear in GROUP BY clause, should they?
</blockquote></details>

Good find, but as I just learned :-) it's enough to group by `route_id`, since it is the primary key of `internal_route.route`. Therefore all other columns from `internal_route.route` can be selected without separately adding them into the `GROUP BY`clause.
___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 417 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlAptVRASa1W8Z0WFSy:-MlHEop74W7_d8vVjVhI:b-rf1v5s) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000500_initial_public_transport/up.sql#L417)):*

> ```
> CREATE INDEX
> ```

Maybe this should be `CREATE UNIQUE INDEX`?
</blockquote></details>

I think not, same situation as above. See comment "The primary key constraint handles the other multicolumn index."

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4)* status: 5 of 18 files reviewed, 28 unresolved discussions (waiting on @datarttu, @jarkkoka, @jschummer, @lassemaatta, @ossiberg, and @pallost)

___
*[hasura/migrations/1620000000500_initial_public_transport/up.sql, line 33 at r18](https://reviewable.io/reviews/hsldevcom/jore4-hasura/4#-MlAeHdt43vVlq6kqbbm:-MlKzNmH-sbRcHTIrh4k:b69phka) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/17b9740efc7cb434e71c015bc26d22dafa92e6b1/hasura/migrations/1620000000500_initial_public_transport/up.sql#L33)):*

I somewhat disagree in the sense that in my opinion, also prefixes could be separated by underscores (or otherwise). Prominent examples might be "pre" and "post".

That being said, submode might still make more sense here as it's less likely to be confused with a metro ("sub" in some parts of the world). Also, we don't have a "super_mode" or similar. So I did change it :+1: 
</blockquote></details>

Clever reasoning. :+1:

PR #10, opened 2021-06-07
Replace placeholders in SQL schema migrations

haphut:
The PostgreSQL usernames for the other users are stored in Docker or Kubernetes secrets. The service responsible for authorizing other users has to read the usernames from the secrets at runtime. Yet the SQL schema migrations are stored in git and stored in the Docker image before runtime.

This conundrum is solved by the script `./replace-placeholders-in-sql-schema-migrations.sh` which is run in the Docker entrypoint. The script does string interpolation by replacing placeholders in the SQL schema migration files with the contents of the Docker secrets.

This script is independent of Hasura and can be used with another SQL schema migrator, ideally without modifications. The script only requires some POSIX tools.

Fixes HSLdevcom/jore4#372.

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)


lassemaatta (commented):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 11 of 11 files at r6, 8 of 8 files at r7, 11 of 11 files at r8, 7 of 7 files at r9, 6 of 6 files at r10, 1 of 1 files at r11, 1 of 1 files at r12, 1 of 1 files at r13.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: all files reviewed, 3 unresolved discussions (waiting on @haphut)

___
*[docker-entrypoint.sh, line 14 at r13](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbeW8vu9Tc6OVmRmMj-:-MbeW8vv2OQNU_2w_bR7:bfn6679) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/4c4795dbbb47c047410b80483d2d898ed443f859/docker-entrypoint.sh#L14)):*
> ```
> HASURA_GRAPHQL_MIGRATIONS_DIR
> ```

For a moment there I thought that this was undefined, but I guess this comes from the environment. Would it be better if this was defined using a similar syntax as the other constants?
___
*[README.md, line 169 at r13](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbeWpjZ1fcN_Aukaz6-:-MbeWpjZ1fcN_Aukaz60:b-jhq3ok) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/4c4795dbbb47c047410b80483d2d898ed443f859/README.md#L169)):*
> ```Markdown
> The mangling is used because of the limitations of SQL identifiers as per [the PostgreSQL documentation](https://www.postgresql.org/docs/13/sql-syntax-lexical.html):
> 
> > SQL identifiers and key words must begin with a letter (a-z, but also letters with diacritical marks and non-Latin letters) or an underscore (\_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($). Note that dollar signs are not allowed in identifiers according to the letter of the SQL standard, so their use might render applications less portable. The SQL standard will not define a key word that contains digits or starts or ends with an underscore, so identifiers of this form are safe against possible conflict with future extensions of the standard.
> ```

I believe there are some limitations on the length of the identifiers in postgres, is that taken into account somewhere?
___
*[replace-placeholders-in-sql-schema-migrations.sh, line 54 at r13](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-Mbeep9Z3OffSdS3iioz:-Mbeep9Z3OffSdS3iip-:b-1xl4vs) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/4c4795dbbb47c047410b80483d2d898ed443f859/replace-placeholders-in-sql-schema-migrations.sh#L54)):*
> ```
>     -q \
>     -x \
> ```

I wonder if it might be a good idea to use the `--long` versions of the options in these scripts (if availlable)? I don't have enough energy to find the correct man page for the commands (for linux, posix, busybox?) to verify what these actually mean :)

haphut (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: all files reviewed, 3 unresolved discussions (waiting on @lassemaatta)

___
*[docker-entrypoint.sh, line 14 at r13](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbeW8vu9Tc6OVmRmMj-:-Mbela3l6kNWRbfncRIk:b8e7mmz) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/4c4795dbbb47c047410b80483d2d898ed443f859/docker-entrypoint.sh#L14)):*
<details><summary><i>Previously, haphut wrote…</i></summary><blockquote>

That is a fair point.
</blockquote></details>

Oops, I pressed the button too early. I will either add a comment or a redundant `FOO=FOO` similar to what is above.

haphut (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: all files reviewed, 3 unresolved discussions (waiting on @lassemaatta)

___
*[replace-placeholders-in-sql-schema-migrations.sh, line 54 at r13](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-Mbeep9Z3OffSdS3iioz:-MbengZ-AkcJ_O-bHWzP:bwhpdb9) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/4c4795dbbb47c047410b80483d2d898ed443f859/replace-placeholders-in-sql-schema-migrations.sh#L54)):*
<details><summary><i>Previously, haphut wrote…</i></summary><blockquote>

I agree, I usually prefer the long options. They are not posix, though. I did not test whether `ash` has them. I will.
</blockquote></details>

Whether Busybox supports them, I mean. The conclusion is that only `sed -f foo -i` can be changed to `sed --file=foo --in-place`. `grep`, `read`, `sort`, `tr`, `rm` did not support long form. I would vote to keep `sed` in short form, as well.

lassemaatta (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: all files reviewed, 2 unresolved discussions (waiting on @lassemaatta)

___
*[replace-placeholders-in-sql-schema-migrations.sh, line 54 at r13](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-Mbeep9Z3OffSdS3iioz:-MbeoouZ6_7TBGTad5ZX:b2nhll9) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/4c4795dbbb47c047410b80483d2d898ed443f859/replace-placeholders-in-sql-schema-migrations.sh#L54)):*
<details><summary><i>Previously, haphut wrote…</i></summary><blockquote>

Whether Busybox supports them, I mean. The conclusion is that only `sed -f foo -i` can be changed to `sed --file=foo --in-place`. `grep`, `read`, `sort`, `tr`, `rm` did not support long form. I would vote to keep `sed` in short form, as well.
</blockquote></details>

Ok.

haphut (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: all files reviewed, 2 unresolved discussions (waiting on @lassemaatta)

___
*[README.md, line 169 at r13](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbeWpjZ1fcN_Aukaz6-:-MbeqRHt6C9WqA1qfAga:b-k0fyb3) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/4c4795dbbb47c047410b80483d2d898ed443f859/README.md#L169)):*
<details><summary><i>Previously, haphut wrote…</i></summary><blockquote>

Not at all! That's a fair point.

https://www.postgresql.org/docs/current/runtime-config-preset.html says
> max_identifier_length (integer)
>
> Reports the maximum identifier length. It is determined as one less than the value of NAMEDATALEN when building the server. The default value of NAMEDATALEN is 64; therefore the default max_identifier_length is 63 bytes, which can be less than 63 characters when using multibyte encodings.

I prefer that to be enforced but documentation is okay too.
</blockquote></details>

I'm not sure what to enforce here. If we hit the limit, would we not just either shorten the name or increase the PostgreSQL limit?

If we end up changing the PostgreSQL option, it becomes tricky for the SQL schema migrator to figure out what is the maximum limit. Maybe the migrator would first connect and query the DB. Or the option value would be stored in another Kubernetes secret.

I'll document that with default PostgreSQL settings and the current bracket delimiter choice of `xxx_` and `_xxx`, the names of the secrets containing the Jore4 DB usernames should be at most 55 ASCII characters and the content of the secret should be at most 63 ASCII characters.

Thanks for bringing this up.

haphut (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: 0 of 51 files reviewed, 2 unresolved discussions (waiting on @lassemaatta)

___
*[docker-entrypoint.sh, line 14 at r13](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbeW8vu9Tc6OVmRmMj-:-MbeuD33Eh-H0yDqAcgY:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/4c4795dbbb47c047410b80483d2d898ed443f859/docker-entrypoint.sh#L14)):*
<details><summary><i>Previously, haphut wrote…</i></summary><blockquote>

Oops, I pressed the button too early. I will either add a comment or a redundant `FOO=FOO` similar to what is above.
</blockquote></details>

Done.
___
*[README.md, line 169 at r13](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbeWpjZ1fcN_Aukaz6-:-MbeuAgw76vtmCrceGov:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/4c4795dbbb47c047410b80483d2d898ed443f859/README.md#L169)):*
<details><summary><i>Previously, haphut wrote…</i></summary><blockquote>

I'm not sure what to enforce here. If we hit the limit, would we not just either shorten the name or increase the PostgreSQL limit?

If we end up changing the PostgreSQL option, it becomes tricky for the SQL schema migrator to figure out what is the maximum limit. Maybe the migrator would first connect and query the DB. Or the option value would be stored in another Kubernetes secret.

I'll document that with default PostgreSQL settings and the current bracket delimiter choice of `xxx_` and `_xxx`, the names of the secrets containing the Jore4 DB usernames should be at most 55 ASCII characters and the content of the secret should be at most 63 ASCII characters.

Thanks for bringing this up.
</blockquote></details>

Done.

lassemaatta (commented):
> Reviewed 1 of 51 files at r14, 1 of 1 files at r15, 1 of 1 files at r16, 1 of 1 files at r17, 1 of 1 files at r18, 11 of 11 files at r19, 8 of 8 files at r20, 11 of 11 files at r21, 7 of 7 files at r22, 6 of 6 files at r23, 1 of 1 files at r24, 1 of 1 files at r25, 1 of 1 files at r26.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @haphut)

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: all files reviewed, 2 unresolved discussions (waiting on @haphut)

___
*[README.md, line 195 at r26](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbfdO8N8LBXXMnw6FuS:-MbfdO8OEo-sl90yNxBH:bzgqt8c) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0292f55dacf83a2342a279ea06bb135a1b9e0d57/README.md#L195)):*
> ```Markdown
> 1. A trivial `sed` oneliner is not enough.
> 1. Use only valid SQL identifier letters in the bracket delimiters.
> 1. `sed`, `grep` and `awk` are available, `perl` and `python` are not.
> ```

I wonder, couldn't we have just installed some tools (like perl or python) to this Hasura docker image instead of trying to save the world with sed, grep and awk? 
___
*[replace-placeholders-in-sql-schema-migrations.sh, line 90 at r26](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjB6SY8tcKSEMpScAk:-MbjB6SZ-9TVsB_qjAlh:b-k91xz0) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0292f55dacf83a2342a279ea06bb135a1b9e0d57/replace-placeholders-in-sql-schema-migrations.sh#L90)):*
> ```Shell
> The migration file
> ${migration_path}
> contains some placeholders for which there are no corresponding secrets. The
> ```

good that you are also checking this case!

lassemaatta (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: all files reviewed, 2 unresolved discussions (waiting on @haphut and @pallost)

___
*[README.md, line 195 at r26](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbfdO8N8LBXXMnw6FuS:-MbjBJm86iYKvnFIxMaU:b-q382xi) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0292f55dacf83a2342a279ea06bb135a1b9e0d57/README.md#L195)):*

I wonder, couldn't we have just installed some tools (like perl or python) to this Hasura docker image instead of trying to save the world with sed, grep and awk? 
</blockquote></details>

The same idea came to me last night when I was thinking about this :) I don't mind a bit of bloat in the image if we can leverage that to make the scripts easier to comprehend.

(But as the code and the tests are already written, I wouldn't necessarily demand that this must be changed immediately)

lassemaatta (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: all files reviewed, 3 unresolved discussions (waiting on @haphut and @pallost)

___
*[README.md, line 167 at r26](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjD-9M5Rwjm4tC4iNF:-MbjD-9M5Rwjm4tC4iNG:b-98hw3d) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0292f55dacf83a2342a279ea06bb135a1b9e0d57/README.md#L167)):*
> ```
> 55 ASCII characters
> ```

Where does the value 55 come from? E.g. why is it not 63?

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: all files reviewed, 5 unresolved discussions (waiting on @haphut and @pallost)

___
*[replace-placeholders-in-sql-schema-migrations.sh, line 32 at r26](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjCdFM29GXTUFkETSe:-MbjCdFM29GXTUFkETSf:bjfhkqw) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0292f55dacf83a2342a279ea06bb135a1b9e0d57/replace-placeholders-in-sql-schema-migrations.sh#L32)):*
> ```Shell
> 
> cleanup() {
>   [ -z ${check_duplicate_list+dummy} ] || rm -f "${check_duplicate_list}"
> ```

A linter would complain that variables should always be in quotes:
`-z "${...}"`
___
*[replace-placeholders-in-sql-schema-migrations.sh, line 83 at r26](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjD1sR5jnIG_2U5SB0:-MbjD1sR5jnIG_2U5SB1:b-g8ymsh) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0292f55dacf83a2342a279ea06bb135a1b9e0d57/replace-placeholders-in-sql-schema-migrations.sh#L83)):*
> ```Shell
>       -h \
>       -E \
>       -e "${start_delimiter}[${placeholder_characters}]+${end_delimiter}" \
> ```

I think you also need a `?` here after the `+` to make sure it's the shortest string that it matches:
`"${start_delimiter}[${placeholder_characters}]+?${end_delimiter}"`

lassemaatta (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: all files reviewed, 5 unresolved discussions (waiting on @haphut and @pallost)

___
*[README.md, line 167 at r26](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjD-9M5Rwjm4tC4iNF:-MbjE2Nk7fKP484YJGUd:b9rle4) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0292f55dacf83a2342a279ea06bb135a1b9e0d57/README.md#L167)):*
<details><summary><i>Previously, lassemaatta (Lasse Määttä) wrote…</i></summary><blockquote>

> ```
> 55 ASCII characters
> ```

Where does the value 55 come from? E.g. why is it not 63?
</blockquote></details>

Actually, why is there a limit in the first place? We don't transfer the names of the secret files to postgres, or do we? Or has this limit something to do with the shell tools we use?

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: all files reviewed, 6 unresolved discussions (waiting on @haphut and @pallost)

___
*[README.md, line 187 at r14](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjEinx8NetNWTPYELn:-MbjEinx8NetNWTPYELo:balkadc) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/df57712e81c9ea202a5f6d187759217caa7a8786/README.md#L187)):*
> ```Markdown
> 1. Fail when a placeholder cannot be replaced.
> 1. Before replacing the placeholders, the SQL schema files must be valid SQL.
> 1. Use the tools available in our Hasura Docker image.
> ```

might be obvious, but could mention that the secret name should not contain `xxx` :)

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: all files reviewed, 7 unresolved discussions (waiting on @haphut and @pallost)

___
*[README.md, line 199 at r14](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjF1h4ADuehPq64AKa:-MbjF1h4ADuehPq64AKb:b-r0w47w) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/df57712e81c9ea202a5f6d187759217caa7a8786/README.md#L199)):*
> ```Markdown
> #### Tests
> 
> There are tests for the string interpolation script in [`./test/string-interpolation`](./test/string-interpolation).
> ```

I would like to see a test with an actual SQL file and syntax

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: all files reviewed, 7 unresolved discussions (waiting on @haphut and @pallost)

___
*[README.md, line 199 at r14](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjF1h4ADuehPq64AKa:-MbjFAXz703RDtvjKLm3:b-h357c2) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/df57712e81c9ea202a5f6d187759217caa7a8786/README.md#L199)):*

I would like to see a test with an actual SQL file and syntax
</blockquote></details>

...with reasonable names for secrets :)

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: all files reviewed, 7 unresolved discussions (waiting on @haphut and @pallost)

___
*[README.md, line 199 at r14](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjF1h4ADuehPq64AKa:-MbjFYng-zoIIfGXy3c7:bec183m) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/df57712e81c9ea202a5f6d187759217caa7a8786/README.md#L199)):*

...with reasonable names for secrets :)
</blockquote></details>

Also, in one of the tests, you are checking that having multiple `xxx_xyz_xxx` placeholders actually do get all replaced. This is a good test, but should have the name a bit more clear to clarify what you are checking there. E.g. `xxx_multiple_placeholders_xxx`

haphut (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: all files reviewed, 7 unresolved discussions (waiting on @haphut, @lassemaatta, and @pallost)

___
*[README.md, line 187 at r14](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjEinx8NetNWTPYELn:-MbkJOmL9rLZx4SCL3Wi:b-mwfkld) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/df57712e81c9ea202a5f6d187759217caa7a8786/README.md#L187)):*

might be obvious, but could mention that the secret name should not contain `xxx` :)
</blockquote></details>

Thanks! I missed that completely. :)
___
*[README.md, line 199 at r14](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjF1h4ADuehPq64AKa:-Mbl3s3b5EGxs4GTSUSr:by13uo3) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/df57712e81c9ea202a5f6d187759217caa7a8786/README.md#L199)):*

Also, in one of the tests, you are checking that having multiple `xxx_xyz_xxx` placeholders actually do get all replaced. This is a good test, but should have the name a bit more clear to clarify what you are checking there. E.g. `xxx_multiple_placeholders_xxx`
</blockquote></details>

Thanks! I'll write a more realistic test case.

I'm not sure what you mean with the last comment, though. The test names come from the names of the directories, here `several-placeholders-per-line/migrations`. Do you mean to also name the placeholders like `xxx_several_xxx xxx_placeholders_xxx xxx_per_xxx xxx_line_xxx`?
___
*[README.md, line 167 at r26](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjD-9M5Rwjm4tC4iNF:-MbkILhf5syvArUggcS2:b52mhsa) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0292f55dacf83a2342a279ea06bb135a1b9e0d57/README.md#L167)):*
<details><summary><i>Previously, lassemaatta (Lasse Määttä) wrote…</i></summary><blockquote>

Actually, why is there a limit in the first place? We don't transfer the names of the secret files to postgres, or do we? Or has this limit something to do with the shell tools we use?
</blockquote></details>

The limit just keeps the placeholders as valid SQL identifiers. Though I doubt that there are static analysis tools that would choke on long identifiers. When we need to run the schema files onto a DB, we'd replace the placeholders, so for that kind of testing the limit is not a problem, either.

Hmm... I will write about the significance or insignificance of following this limit.

`63 - length("xxx_") - length("_xxx") == 55`
___
*[README.md, line 195 at r26](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbfdO8N8LBXXMnw6FuS:-MbkJfWiELZGZoC2ssME:b6m90t6) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0292f55dacf83a2342a279ea06bb135a1b9e0d57/README.md#L195)):*
<details><summary><i>Previously, lassemaatta (Lasse Määttä) wrote…</i></summary><blockquote>

The same idea came to me last night when I was thinking about this :) I don't mind a bit of bloat in the image if we can leverage that to make the scripts easier to comprehend.

(But as the code and the tests are already written, I wouldn't necessarily demand that this must be changed immediately)
</blockquote></details>

Yes, that does sound lucrative. In advance I thought this would be a shorter endeavour.

I believe this script and its tests can be used as such with the next SQL migrator as long as its Docker image has shell. Though for most images it's easy to install e.g. python if it's missing. We likely will create our own Dockerfile for the next SQL migrator, anyway. So a bit of a moot point.

To keep the files as valid SQL even with the placeholders, the bracket delimiters need to consist of `[0-9A-Za-z_]`. Therefore most existing templating libraries with their own template languages do not fit. So while with python or typescript we'd get more ergonomic code to read and write, we'd end up writing pretty much the same functionality, apart from probably using `split()` instead of regex.

I also thought that the testing will be just a few lines. Here a test framework that makes it easy to handle exit codes and stdout and stderr from a shell process would have been handy, in hindsight.

I'm okay with losing the sunk cost. The cost ahead matters, though, as @lassemaatta points out. We're I to start again, I'd follow your advice.

haphut:
___
*[README.md, line 195 at r26](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbfdO8N8LBXXMnw6FuS:-Mbl6Ri705ZbRmKLzybP:bb03j39) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0292f55dacf83a2342a279ea06bb135a1b9e0d57/README.md#L195)):*
<details><summary><i>Previously, haphut wrote…</i></summary><blockquote>

Yes, that does sound lucrative. In advance I thought this would be a shorter endeavour.

I believe this script and its tests can be used as such with the next SQL migrator as long as its Docker image has shell. Though for most images it's easy to install e.g. python if it's missing. We likely will create our own Dockerfile for the next SQL migrator, anyway. So a bit of a moot point.

To keep the files as valid SQL even with the placeholders, the bracket delimiters need to consist of `[0-9A-Za-z_]`. Therefore most existing templating libraries with their own template languages do not fit. So while with python or typescript we'd get more ergonomic code to read and write, we'd end up writing pretty much the same functionality, apart from probably using `split()` instead of regex.

I also thought that the testing will be just a few lines. Here a test framework that makes it easy to handle exit codes and stdout and stderr from a shell process would have been handy, in hindsight.

I'm okay with losing the sunk cost. The cost ahead matters, though, as @lassemaatta points out. We're I to start again, I'd follow your advice.
</blockquote></details>

If you think this should be rewritten, it shall be done.

haphut (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: 0 of 72 files reviewed, 7 unresolved discussions (waiting on @haphut, @lassemaatta, and @pallost)

___
*[README.md, line 187 at r14](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjEinx8NetNWTPYELn:-MchcWgOEvFIiyoFxAeb:bogny6h) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/df57712e81c9ea202a5f6d187759217caa7a8786/README.md#L187)):*
<details><summary><i>Previously, haphut wrote…</i></summary><blockquote>

Thanks! I missed that completely. :)
</blockquote></details>

The script now checks for this case and gives an error if a secret name includes `xxx`.
___
*[README.md, line 199 at r14](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjF1h4ADuehPq64AKa:-MchdPO-5O7TfjpZtGqV:bepfyl3) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/df57712e81c9ea202a5f6d187759217caa7a8786/README.md#L199)):*
<details><summary><i>Previously, haphut wrote…</i></summary><blockquote>

Thanks! I'll write a more realistic test case.

I'm not sure what you mean with the last comment, though. The test names come from the names of the directories, here `several-placeholders-per-line/migrations`. Do you mean to also name the placeholders like `xxx_several_xxx xxx_placeholders_xxx xxx_per_xxx xxx_line_xxx`?
</blockquote></details>

There's a more realistic test case, as well.

I'm still not sure how you would like to change the test `several-placeholders-per-line`.
___
*[README.md, line 167 at r26](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjD-9M5Rwjm4tC4iNF:-MchcSli2NxGqKQIf8b8:b-xserin) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0292f55dacf83a2342a279ea06bb135a1b9e0d57/README.md#L167)):*
<details><summary><i>Previously, haphut wrote…</i></summary><blockquote>

The limit just keeps the placeholders as valid SQL identifiers. Though I doubt that there are static analysis tools that would choke on long identifiers. When we need to run the schema files onto a DB, we'd replace the placeholders, so for that kind of testing the limit is not a problem, either.

Hmm... I will write about the significance or insignificance of following this limit.

`63 - length("xxx_") - length("_xxx") == 55`
</blockquote></details>

This is now documented.
___
*[replace-placeholders-in-sql-schema-migrations.sh, line 32 at r26](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjCdFM29GXTUFkETSe:-Mchdc-vA4f9BFarfUik:bt2vr1b) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0292f55dacf83a2342a279ea06bb135a1b9e0d57/replace-placeholders-in-sql-schema-migrations.sh#L32)):*

A linter would complain that variables should always be in quotes:
`-z "${...}"`
</blockquote></details>

Fixed.
___
*[replace-placeholders-in-sql-schema-migrations.sh, line 83 at r26](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbjD1sR5jnIG_2U5SB0:-MchfMEa-t0qraRpqhzb:bun4rp1) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0292f55dacf83a2342a279ea06bb135a1b9e0d57/replace-placeholders-in-sql-schema-migrations.sh#L83)):*

I think you also need a `?` here after the `+` to make sure it's the shortest string that it matches:
`"${start_delimiter}[${placeholder_characters}]+?${end_delimiter}"`
</blockquote></details>

Thanks! That's a good point.

The lazy modifier `?` requires PCRE which is not available for this `grep` implementation. Adding `?` there makes the regex match with `xxx__xxx`. As there cannot be a secret filename of length 0, the approach fails the existing tests.

Similarly using a negative lookahead for `_xxx` would require PCRE.

I cannot think of a fix for this with the available shell utilities.

Fortunately this bug only seems to affect the misleading error message regarding catenated leftover placeholders.

E.g. if both secrets `foo` and `bar` are missing, then using two placeholders `xxx_foo_xxxxxx_bar_xxx` in the SQL would cause an error message complaining about only one placeholder. That could mislead the developer into adding a secret named `foo-xxxxxx-bar` which would fail on the next iteration because of the substring `xxx`. The developer would by then hopefully notice that two secrets are required.

pallost (dismissed):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: 0 of 72 files reviewed, 9 unresolved discussions (waiting on @haphut, @lassemaatta, and @pallost)

___
*[Dockerfile, line 3 at r42](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-Mci6EyQ6M52nHTn0WMH:-Mci6EyQ6M52nHTn0WMI:b-605kzw) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/aecd917d583b95d6ae94847e9c1911634789370f/Dockerfile#L3)):*
> ```Dockerfile
> FROM hasura/graphql-engine:v1.3.3.cli-migrations-v2
> ENV HASURA_GRAPHQL_UNAUTHORIZED_ROLE="anonymous"
> ENV HASURA_GRAPHQL_MIGRATIONS_DIR="/hasura-migrations"
> ```

should also define the `HASURA_GRAPHQL_METADATA_DIR` the same way as it is static aswell
___
*[Dockerfile, line 7 at r42](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-Mci5s2K8bPhdb80aCbS:-Mci5s2K8bPhdb80aCbT:bc7elm) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/aecd917d583b95d6ae94847e9c1911634789370f/Dockerfile#L7)):*
> ```Dockerfile
> COPY ./hasura/metadata /hasura-metadata
> WORKDIR /app
> COPY ./replace-placeholders-in-sql-schema-migrations.sh ./
> ```

it would be nicer if `replace-placeholders-in-sql-schema-migrations.sh` and `docker-entrypoint.sh` would be stored in a `scripts` directory in the repository
___
*[docker-entrypoint.sh, line 7 at r42](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-Mci6UwU2aThUuP1aWSy:-Mci6UwU2aThUuP1aWSz:b-x9ag1z) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/aecd917d583b95d6ae94847e9c1911634789370f/docker-entrypoint.sh#L7)):*
> ```Shell
> SECRET_STORE_BASE_PATH="${SECRET_STORE_BASE_PATH:-/run/secrets}"
> HASURA_ADMIN_SECRET="$(cat "${SECRET_STORE_BASE_PATH}/hasura-admin-secret")"
> DB_USERNAME="$(cat "${SECRET_STORE_BASE_PATH}/db-username")"
> ```

in the future, you could use the `read-secrets.sh` from https://github.com/HSLdevcom/jore4-tools
___
*[README.md, line 195 at r26](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-MbfdO8N8LBXXMnw6FuS:-Mci72kK4_Li3hd1oNgW:b-8ch0vx) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0292f55dacf83a2342a279ea06bb135a1b9e0d57/README.md#L195)):*
<details><summary><i>Previously, haphut wrote…</i></summary><blockquote>

If you think this should be rewritten, it shall be done.
</blockquote></details>

well, let's not throw it away, the important thing is that this works :)

haphut (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: 0 of 73 files reviewed, 9 unresolved discussions (waiting on @haphut, @lassemaatta, and @pallost)

___
*[Dockerfile, line 3 at r42](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-Mci6EyQ6M52nHTn0WMH:-MciVp9t8-3UjphaeZsX:bb6un1j) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/aecd917d583b95d6ae94847e9c1911634789370f/Dockerfile#L3)):*

should also define the `HASURA_GRAPHQL_METADATA_DIR` the same way as it is static aswell
</blockquote></details>

Done
___
*[Dockerfile, line 7 at r42](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-Mci5s2K8bPhdb80aCbS:-MciVqXC8iMjxtfEqrVc:bb6un1j) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/aecd917d583b95d6ae94847e9c1911634789370f/Dockerfile#L7)):*

it would be nicer if `replace-placeholders-in-sql-schema-migrations.sh` and `docker-entrypoint.sh` would be stored in a `scripts` directory in the repository
</blockquote></details>

Done
___
*[docker-entrypoint.sh, line 7 at r42](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10#-Mci6UwU2aThUuP1aWSy:-MciVtw5DX5_rNXZ1wwp:beeq3h0) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/aecd917d583b95d6ae94847e9c1911634789370f/docker-entrypoint.sh#L7)):*

in the future, you could use the `read-secrets.sh` from https://github.com/HSLdevcom/jore4-tools
</blockquote></details>

Acknowledged! But not done.

lassemaatta (dismissed):
> Reviewed 1 of 70 files at r43, 1 of 1 files at r46, 11 of 11 files at r48, 6 of 6 files at r49, 11 of 11 files at r50, 7 of 7 files at r51, 6 of 6 files at r52, 6 of 6 files at r53, 8 of 8 files at r54, 7 of 7 files at r55, 1 of 1 files at r58, 7 of 7 files at r59.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: 71 of 73 files reviewed, 8 unresolved discussions (waiting on @haphut, @lassemaatta, and @pallost)

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/10)* status: 0 of 74 files reviewed, all discussions resolved (waiting on @lassemaatta)

PR #42, opened 2022-01-11
Scheduled stop point infra link constraints

jschummer:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42)


jarkkoka (commented):
Very complicated code. Had a two hour session with it. Could read just one half of it and not with confidence. Need to continue later. Maybe needs some other peeps too to have a look at this.

Test coverage is good which certainly is a plus.

> Reviewed 39 of 55 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-:-MtO-FdADqgULl2s3PSS:b-4wv6lc)* status: 39 of 55 files reviewed, 6 unresolved discussions (waiting on @jarkkoka and @jschummer)

___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts, line 55 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNqluB9_0ufXF0lfRU:-MtNqluB9_0ufXF0lfRV:b-f5mwna) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts#L55)):*
> ```XML
>         .then(
>           checkErrorResponse(
>             "found stop in journey pattern which is on a link that is not part of the route"
> ```

Wondering whether this error message is really correct? From the test fixture, it seems to me that the link IS A PART of the route.



_Code quote:_
```XML
"found stop in journey pattern which is on a link that is not part of the route"
```
___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts, line 116 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNs0eRBl3QB9eR4yIS:-MtNs0eSBAgHClzEjKZr:b-at2te2) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts#L116)):*
> ```XML
>             toBeRemoved.route_id,
>             toBeRemoved.infrastructure_link_id
>           ),
> ```

This could be assigned to named function variable since it is repeated.



_Code quote:_
```XML
      await rp.post({
        ...config.hasuraRequestTemplate,
        body: {
          query: createMutation(
            toBeRemoved.route_id,
            toBeRemoved.infrastructure_link_id
          ),
        },
      });
```
___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts, line 95 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNtNF08eiqGjre9tKw:-MtNtNF08eiqGjre9tKx:b-h6gob7) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts#L95)):*
> ```XML
>     const toBeInserted = createToBeInserted(
>       journeyPatterns[0].journey_pattern_id,
>       scheduledStopPoints[3].scheduled_stop_point_id,
> ```

As a reviewer, it feels too tedious to check that the invariant/assumption really holds. I will just trust that you have been careful in setting array indices correctly.



_Code quote:_
```XML
      journeyPatterns[0].journey_pattern_id,
      scheduledStopPoints[3].scheduled_stop_point_id,
```
___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts, line 122 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNt7ku8gxUsx_WMNl3:-MtNt7ku8gxUsx_WMNl4:b-txdou6) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts#L122)):*
> ```XML
>   });
> 
>   describe("when stop's link is traversed in a direction not compatible with the stop's direction", () => {
> ```

matching



_Code quote:_
```XML
compatible
```
___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts, line 140 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNu3t4BDFhIy7dE9WI:-MtNu3t4BDFhIy7dE9WJ:bw5tisy) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts#L140)):*
> ```XML
>     const toBeInserted = createToBeInserted(
>       journeyPatterns[0].journey_pattern_id,
>       scheduledStopPoints[1].scheduled_stop_point_id,
> ```

Maybe generation of matching/mismatching pair should be very near to the code that generates the test data. In Reviewable system, it is quite laborous to check that this test fixture is properly set up. I am not mandating a change here, however. I'm just making a point that testing is in this form is quite fragile. For instance, if you later make changes or augment the test data it is pretty difficult to spot all the referencing sites that should be updated to new changes. Luckily you have raised test coverage to such a high level, that it is probably sufficient to just trust that existing test cases will break if data is changed to be incompatible with them. 



_Code quote:_
```XML
      journeyPatterns[0].journey_pattern_id,
      scheduledStopPoints[1].scheduled_stop_point_id,
```
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 27 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNzqyk-TWJ7zxuf6e5:-MtNzqyk-TWJ7zxuf6e6:b-dub733) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L27)):*
> ```TSQL
>                                      AS route_order
>   FROM journey_pattern.scheduled_stop_point_in_journey_pattern sspijp
>          LEFT JOIN internal_service_pattern.scheduled_stop_point ssp
> ```

This whole function would need a lot of explaining comments. I cannot read this without comments.



_Code quote:_
```TSQL
  CREATE TEMP TABLE infra_link_map AS
  SELECT sspijp.journey_pattern_id,
         sspijp.scheduled_stop_point_id,
         sspijp.scheduled_stop_point_sequence,
         row_number()
         OVER (PARTITION BY sspijp.journey_pattern_id ORDER BY sspijp.scheduled_stop_point_sequence)
                                     AS journey_pattern_order,
         ssp.direction               AS scheduled_stop_point_direction,
         ilar.route_id,
         ilar.infrastructure_link_id AS route_link_id,
         ilar.infrastructure_link_sequence,
         ilar.is_traversal_forwards  AS is_link_traversal_forwards,
         -- order also by scheduled_stop_point_sequence to get a deterministic order also when there are multiple stops on the same infra link
         row_number()
         OVER (PARTITION BY sspijp.journey_pattern_id ORDER BY ilar.infrastructure_link_sequence, sspijp.scheduled_stop_point_sequence)
                                     AS route_order
  FROM journey_pattern.scheduled_stop_point_in_journey_pattern sspijp
         LEFT JOIN internal_service_pattern.scheduled_stop_point ssp
                   ON ssp.scheduled_stop_point_id = sspijp.scheduled_stop_point_id
         LEFT JOIN journey_pattern.journey_pattern jp ON jp.journey_pattern_id = sspijp.journey_pattern_id
         LEFT JOIN route.infrastructure_link_along_route ilar
                   ON ilar.route_id = jp.on_route_id AND
                      ilar.infrastructure_link_id = ssp.located_on_infrastructure_link_id
  WHERE sspijp.journey_pattern_id = filter_journey_pattern_id
     OR jp.on_route_id = filter_route_id;
```

jschummer (commented):
Thank you very much for taking this on. I think @pkainulainen-vincit already had a look at the part of this PR, which didn't have the integration tests included yet and he didn't come up with anything wrong (except he also wanted someone else to take a look).

I think it's quite good already if you can focus on the actual constraint, I'm not sure you necessarily need to check each and every line of the test code. But then that's just my humble opinion, of course you're free to check the test code too if you want.

More comments below.

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-:-Mtaz1njEogwDTlxY12B:bxz540)* status: 38 of 55 files reviewed, 6 unresolved discussions (waiting on @jarkkoka)

___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts, line 55 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNqluB9_0ufXF0lfRU:-Mtb0MdT5qVDAZA33bB9:bkomas1) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts#L55)):*

Wondering whether this error message is really correct? From the test fixture, it seems to me that the link IS A PART of the route.
</blockquote></details>

Mmmh, how did you think that?

The used mutation should (try to) remove the infrastructure_link_along_route row, which ties the given link to the route. So the constraint trigger should catch this situation (and apparently we do get the right error message in the response).
___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts, line 116 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNs0eRBl3QB9eR4yIS:-MtaztGw8Y_tbBqC0N4g:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts#L116)):*

This could be assigned to named function variable since it is repeated.
</blockquote></details>

Done.
___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts, line 95 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNtNF08eiqGjre9tKw:-MtazuxW8MjbPhR8qiJt:bbwt73a) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts#L95)):*

As a reviewer, it feels too tedious to check that the invariant/assumption really holds. I will just trust that you have been careful in setting array indices correctly.
</blockquote></details>

Yes, you have a good point, it's not easy to follow these, by any means. After the test refactoring commit (included), it would be rather easy to have separate test sets for each test, but I'm not sure how much it would easy reviewing the code (since you still need to look at many files to see the relationships between stop points and infra links). But it might be an option for future tests, since it could ease writing the tests.
___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts, line 122 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNt7ku8gxUsx_WMNl3:-Mtb-XXv1UnwWgXNb4xz:bh0yr6y) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts#L122)):*

matching
</blockquote></details>

I didn't want to use "matching", because it sounds to much like "equals", which does not hold. If the stop's direction is "bidirectional", the link can be traversed "forwards" or "backwards".

I suppose "matching" could be the correct term, but I'm afraid it could easily be interpreted the wrong way. Can change it though if you feel it's better.
___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts, line 140 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNu3t4BDFhIy7dE9WI:-Mtb-w-0-H_WEAI0tRmB:b-ozjl8z) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts#L140)):*

Maybe generation of matching/mismatching pair should be very near to the code that generates the test data. In Reviewable system, it is quite laborous to check that this test fixture is properly set up. I am not mandating a change here, however. I'm just making a point that testing is in this form is quite fragile. For instance, if you later make changes or augment the test data it is pretty difficult to spot all the referencing sites that should be updated to new changes. Luckily you have raised test coverage to such a high level, that it is probably sufficient to just trust that existing test cases will break if data is changed to be incompatible with them. 
</blockquote></details>

Like mentioned above, after the test refactoring commit (included), it would be rather easy to have separate test sets for each test, but I'm not sure how much it would easy reviewing the code (since you still need to look at many files to see the relationships between stop points and infra links). But it might be an option for future tests, since it could ease writing the tests.

And I think splitting the test data sets is probably the way to go if some future change should cause a (non-trivial) breakage of the tests.
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 27 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNzqyk-TWJ7zxuf6e5:-Mtb0CVn-B3y0YMny0O6:b26kf0v) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L27)):*

This whole function would need a lot of explaining comments. I cannot read this without comments.
</blockquote></details>

:+1: I added some comments now, hopefully it's now easier to grasp what is going on.

jarkkoka (changes requested):
I think I found a bug in constraint but I'm not 100% sure. Anyways, props for working hard to keep data integrity at a high level!

> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-:-Mtc7-F31U1AVcJL6d8a:b-mn01lp)* status: 40 of 55 files reviewed, 7 unresolved discussions (waiting on @jarkkoka and @jschummer)

___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts, line 55 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNqluB9_0ufXF0lfRU:-MtbkZ_l3segB0TEGwyZ:b-1y2s2p) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts#L55)):*

Mmmh, how did you think that?

The used mutation should (try to) remove the infrastructure_link_along_route row, which ties the given link to the route. So the constraint trigger should catch this situation (and apparently we do get the right error message in the response).
</blockquote></details>

Yes, constraint trigger does the right thing in raising an exception. It is the error message which made me confused.

In this test fixture, the link is originally part of the route but trigger catches the situation after it was removed (and not anymore part of link). I am not sure whether those constraint error messages should be generalised more since one error message may cover many different situations. E.g. this error message might cover other cases as well in addition to removing a link from a route like adding a stop to journey pattern where stop is not along the related route.
___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts, line 122 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNt7ku8gxUsx_WMNl3:-MtbmqNC3j6Z4RJBveHZ:b-nrlofb) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts#L122)):*

I didn't want to use "matching", because it sounds to much like "equals", which does not hold. If the stop's direction is "bidirectional", the link can be traversed "forwards" or "backwards".

I suppose "matching" could be the correct term, but I'm afraid it could easily be interpreted the wrong way. Can change it though if you feel it's better.
</blockquote></details>

I think the verb `match` does not have the same meaning as `equal`. Look at these two Cambridge dictionary examples:
- *The shirt and pants match perfectly.*
-  *Her fingerprints matched the prints that were taken from the crime scene.*

I think my suggestion would fit this case pretty well because match verb can be used to tie together things of different kind that have similarity in some sense. However, let's not argue about non-essential things. If you want to stick to word `compatible` despite of my comment, then you use it.
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 27 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNzqyk-TWJ7zxuf6e5:-MtbpzdIEwsLI_GkjPue:bmauxvd) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L27)):*

:+1: I added some comments now, hopefully it's now easier to grasp what is going on.
</blockquote></details>

Thanks! Good additions. It helped a lot!  :+1: 
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 22 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtbtX6x3_h1nmwE57vF:-MtbtX6x3_h1nmwE57vG:b-clfb6r) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a5251c145776735a0b68718ba0686c0e68e7576a/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L22)):*
> ```TSQL
>          row_number()
>          OVER (PARTITION BY sspijp.journey_pattern_id ORDER BY sspijp.scheduled_stop_point_sequence)
>                                      AS journey_pattern_order,
> ```

Isn't this dynamically computed `journey_pattern_order` exactly the same as `scheduled_stop_point_sequence` that already exists and appears above? If not, why?

The answers to the two previous questions are keys to understand (one part of) this function's logic.

As a reviewer, I think this column should be renamed to `verified_journey_pattern_order` if we cannot trust the value of `scheduled_stop_point_sequence` column at this point. This would make this function more understandable if my assumption is correct.



_Code quote:_
```TSQL
         row_number()
         OVER (PARTITION BY sspijp.journey_pattern_id ORDER BY sspijp.scheduled_stop_point_sequence)
                                     AS journey_pattern_order,
```
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 42 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtbyEgLCFfxT1UvqImx:-MtbyEgLCFfxT1UvqImy:b-g1khav) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a5251c145776735a0b68718ba0686c0e68e7576a/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L42)):*
> ```TSQL
>          LEFT JOIN route.infrastructure_link_along_route ilar
>                    ON ilar.route_id = jp.on_route_id AND
>                       ilar.infrastructure_link_id = ssp.located_on_infrastructure_link_id
> ```

What if one infrastructure link is traversed twice on a route? It seems to me that this would make two joins from an SSP to an infra-link-along-route. I guess this does not work in such scenario. What do you think? Do you agree that there is a logic error in such case?

In Jore3, there exist routes with U-turns. E.g. on route "4433K" one infrastructure link is traversed twice (in different direction) after a U-turn.

Quite likely one single stop (SSP) exists at most once on a journey pattern but I'm not 100% sure.



_Code quote:_
```TSQL
                   ON ilar.route_id = jp.on_route_id AND
                      ilar.infrastructure_link_id = ssp.located_on_infrastructure_link_id
```
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 77 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-Mtc-qGg2QPHxs0_-sS3:-Mtc-qGg2QPHxs0_-sS4:b-gpufx6) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a5251c145776735a0b68718ba0686c0e68e7576a/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L77)):*
> ```TSQL
>   THEN
>     DROP TABLE infra_link_map;
>     RAISE EXCEPTION 'found stop in journey pattern, whose infra link is traversed in a direction not compatible with the stop''s directions';
> ```

an extra quote to be removed



_Code quote:_
```TSQL
'
```
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 141 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-Mtc2E9rCuZDZoQLBheH:-Mtc2E9s2JufuCbowAUd:b-kyrdwu) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a5251c145776735a0b68718ba0686c0e68e7576a/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L141)):*
> ```TSQL
>      - included stop might be included at a position, which does not correspond to the position of its
>        infra link in the journey pattern''s route
>      - included stop''s link might be traversed in a direction not suitable with the route''s traversal
> ```

Again, the word "matching" would be more clear to the reader IMO. :)



_Code quote:_
```TSQL
suitable
```

jschummer (commented):
Mhh yees, I think you might have a point with the link that is traversed twice. I guess I have to write some more tests for that case (with their own data set this time) and let them fail until I found a fix :-)

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-:-MtcCCVZ8zfKuIHzl9KF:b-atwlsc)* status: 40 of 55 files reviewed, 7 unresolved discussions (waiting on @jarkkoka and @jschummer)

___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts, line 55 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNqluB9_0ufXF0lfRU:-MtcC_Q_1m8PpbbevLhT:b-sbrn0b) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts#L55)):*

Yes, constraint trigger does the right thing in raising an exception. It is the error message which made me confused.

In this test fixture, the link is originally part of the route but trigger catches the situation after it was removed (and not anymore part of link). I am not sure whether those constraint error messages should be generalised more since one error message may cover many different situations. E.g. this error message might cover other cases as well in addition to removing a link from a route like adding a stop to journey pattern where stop is not along the related route.
</blockquote></details>

Yes, this is the same error message in both cases. From the point of view of the trigger, it's exactly the same situation and the message describes the situation found by the constraint check. If you want to suggest better wording, I'm open to suggestions.
___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts, line 122 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNt7ku8gxUsx_WMNl3:-MtcD-u52Edrd9w45J9w:b3a4sim) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts#L122)):*

I think the verb `match` does not have the same meaning as `equal`. Look at these two Cambridge dictionary examples:
- *The shirt and pants match perfectly.*
-  *Her fingerprints matched the prints that were taken from the crime scene.*

I think my suggestion would fit this case pretty well because match verb can be used to tie together things of different kind that have similarity in some sense. However, let's not argue about non-essential things. If you want to stick to word `compatible` despite of my comment, then you use it.
</blockquote></details>

Heh, I think the fingerprint example is not necessarily on your side :-)

But I do agree that the "shirt and pants" example clearly is. I'll fix this, shouldn't be too much work.
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 22 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtbtX6x3_h1nmwE57vF:-MtcD_Im2fwX1e9UphP4:bm4s3fo) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a5251c145776735a0b68718ba0686c0e68e7576a/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L22)):*

Isn't this dynamically computed `journey_pattern_order` exactly the same as `scheduled_stop_point_sequence` that already exists and appears above? If not, why?

The answers to the two previous questions are keys to understand (one part of) this function's logic.

As a reviewer, I think this column should be renamed to `verified_journey_pattern_order` if we cannot trust the value of `scheduled_stop_point_sequence` column at this point. This would make this function more understandable if my assumption is correct.
</blockquote></details>

The scheduled_stop_point_sequence does not need to be continuous, i.e. without gaps.

At some point, it was suggested that a view would expose a gapless version of the sequence numbers to the graphql API, but that has the problem that the API might behave confusingly when inserting new stop points. (Imagine the sequence - as seen by the API user - contains elements 0, 1, and 2. Then the API user would like to insert a new stop at position two, so it shifts the stop at 2 to position 3. But the gapless view would still present sequence numbers 0, 1, and 2 after the shift.)

The other option would be a constraint to enforce the sequence to be continuous, but I didn't see that this is necessary, since every "user" of the table (API user or this constraint) can easily reorder the sequence by themselves.

I can add this info as a comment and also rename the column to clarify this.
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 42 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtbyEgLCFfxT1UvqImx:-MtcFGqP8mCBXbDWSTPF:b5upb82) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a5251c145776735a0b68718ba0686c0e68e7576a/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L42)):*

What if one infrastructure link is traversed twice on a route? It seems to me that this would make two joins from an SSP to an infra-link-along-route. I guess this does not work in such scenario. What do you think? Do you agree that there is a logic error in such case?

In Jore3, there exist routes with U-turns. E.g. on route "4433K" one infrastructure link is traversed twice (in different direction) after a U-turn.

Quite likely one single stop (SSP) exists at most once on a journey pattern but I'm not 100% sure.
</blockquote></details>

This is probably (and unfortunately) a very good finding. I guess I have to sit down once more and think this through. And write some failing tests before trying to fix this..
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 77 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-Mtc-qGg2QPHxs0_-sS3:-MtcFU2b6LKJ0DKDU7vS:b4qezs0) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a5251c145776735a0b68718ba0686c0e68e7576a/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L77)):*

an extra quote to be removed
</blockquote></details>

The double quote is needed to escape the single quote. Unless there is another quote somewhere which I don't see right now...
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 141 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-Mtc2E9rCuZDZoQLBheH:-MtcFfq1EhfIPknrZ4_-:bhiubyx) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a5251c145776735a0b68718ba0686c0e68e7576a/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L141)):*

Again, the word "matching" would be more clear to the reader IMO. :)
</blockquote></details>

Ok, ok, I think I'll change it :-)

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42)* status: 40 of 55 files reviewed, 4 unresolved discussions (waiting on @jarkkoka and @jschummer)

___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts, line 55 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNqluB9_0ufXF0lfRU:-MtcK37h0QcUWr3S3Cj-:b-jsnoy1) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts#L55)):*

Yes, this is the same error message in both cases. From the point of view of the trigger, it's exactly the same situation and the message describes the situation found by the constraint check. If you want to suggest better wording, I'm open to suggestions.
</blockquote></details>

The issue with the error message is that it represents an illegal state that we actually will never have (thanks to the constraint). The message should better indicate that the constraint trigger is preventing a transition to illegal state.

Hence, I will suggest: "Cannot transition to state where a stop in journey pattern is located on an infrastructure link that is not part of the route associated with the journey pattern"
___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts, line 122 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNt7ku8gxUsx_WMNl3:-MtcG_9B2gbRsE2AvSTh:b32vhw) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts#L122)):*

Heh, I think the fingerprint example is not necessarily on your side :-)

But I do agree that the "shirt and pants" example clearly is. I'll fix this, shouldn't be too much work.
</blockquote></details>

:)
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 22 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtbtX6x3_h1nmwE57vF:-MtcL6-X7iuQVQCgSyhV:b-14q3c7) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a5251c145776735a0b68718ba0686c0e68e7576a/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L22)):*

The scheduled_stop_point_sequence does not need to be continuous, i.e. without gaps.

At some point, it was suggested that a view would expose a gapless version of the sequence numbers to the graphql API, but that has the problem that the API might behave confusingly when inserting new stop points. (Imagine the sequence - as seen by the API user - contains elements 0, 1, and 2. Then the API user would like to insert a new stop at position two, so it shifts the stop at 2 to position 3. But the gapless view would still present sequence numbers 0, 1, and 2 after the shift.)

The other option would be a constraint to enforce the sequence to be continuous, but I didn't see that this is necessary, since every "user" of the table (API user or this constraint) can easily reorder the sequence by themselves.

I can add this info as a comment and also rename the column to clarify this.
</blockquote></details>

Ok. Thanks for clarification. I did not know those details. I hope you add comments explaining those matters.
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 42 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtbyEgLCFfxT1UvqImx:-MtcHVzO4xjhwFLrvC_H:b55lphj) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a5251c145776735a0b68718ba0686c0e68e7576a/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L42)):*

This is probably (and unfortunately) a very good finding. I guess I have to sit down once more and think this through. And write some failing tests before trying to fix this..
</blockquote></details>

Yes, I admit that it is not easy to fix. Writing tests is probably the only way to approach the problem since the solution is complicated.
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 77 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-Mtc-qGg2QPHxs0_-sS3:-MtcJFNi2lkjpW0dV828:b-ts5ql4) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a5251c145776735a0b68718ba0686c0e68e7576a/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L77)):*

The double quote is needed to escape the single quote. Unless there is another quote somewhere which I don't see right now...
</blockquote></details>

Ok, my bad.

datarttu (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42)* status: 40 of 55 files reviewed, 4 unresolved discussions (waiting on @datarttu, @jarkkoka, and @jschummer)

___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 42 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtbyEgLCFfxT1UvqImx:-MtqmI_cCN_G6NGDmG8K:b-uclpg4) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a5251c145776735a0b68718ba0686c0e68e7576a/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L42)):*

Yes, I admit that it is not easy to fix. Writing tests is probably the only way to approach the problem since the solution is complicated.
</blockquote></details>

Here's a good example of a pretty messy route from real life: https://reittiopas.hsl.fi/linjat/HSL:2116/pysakit/HSL:2116:1:01 (route 2116 in Jore), perhaps worth testing with! It has a lot of loops and dead-ends, and currently seems to be the only route where the same stop is used more than once in the journey pattern (E2108 WeeGee-talo). Still, even one example is enough to prove that also duplicated stops should be considered as a legit edge case, and we might well have such messy routes in the future as well :)

I think it might be a good idea to warn the Jore4 user if the same stop point is used more than once by a journey pattern, though. But maybe that's a different story and need.

jschummer (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42)* status: 35 of 63 files reviewed, 4 unresolved discussions (waiting on @datarttu and @jarkkoka)

___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts, line 55 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNqluB9_0ufXF0lfRU:-Mu9lqKtEVfPGD3tpf8l:b9ud7os) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts#L55)):*

The issue with the error message is that it represents an illegal state that we actually will never have (thanks to the constraint). The message should better indicate that the constraint trigger is preventing a transition to illegal state.

Hence, I will suggest: "Cannot transition to state where a stop in journey pattern is located on an infrastructure link that is not part of the route associated with the journey pattern"
</blockquote></details>

Yes, I do see your point, I guess it was a bit misleading. Maybe the new form ("must match") is clearer?
___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts, line 122 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNt7ku8gxUsx_WMNl3:-MuF-v9v-Uh_0XC1a-hV:bg3pnzi) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts#L122)):*

:)
</blockquote></details>

I now changed the occurrance(s) of "compatible" to "matching" in this PR. There are still some left from previous PRs. If these should be changed too, that would require follow-up work.
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 27 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNzqyk-TWJ7zxuf6e5:-Mu9rx9q5ZClgwCKiqql:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L27)):*

Thanks! Good additions. It helped a lot!  :+1: 
</blockquote></details>

Done.
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 22 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtbtX6x3_h1nmwE57vF:-Mu9rnPb7aQX7gnMFXn_:bsavf4c) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a5251c145776735a0b68718ba0686c0e68e7576a/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L22)):*

Ok. Thanks for clarification. I did not know those details. I hope you add comments explaining those matters.
</blockquote></details>

I also tried to explain the workings of the new versions in comments. Hopefully they are readable.
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 42 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtbyEgLCFfxT1UvqImx:-Mu9ryDQ6ted8Ht0v4rg:b8yc6so) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a5251c145776735a0b68718ba0686c0e68e7576a/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L42)):*
<details><summary><i>Previously, datarttu (Arttu Kosonen) wrote…</i></summary><blockquote>

Here's a good example of a pretty messy route from real life: https://reittiopas.hsl.fi/linjat/HSL:2116/pysakit/HSL:2116:1:01 (route 2116 in Jore), perhaps worth testing with! It has a lot of loops and dead-ends, and currently seems to be the only route where the same stop is used more than once in the journey pattern (E2108 WeeGee-talo). Still, even one example is enough to prove that also duplicated stops should be considered as a legit edge case, and we might well have such messy routes in the future as well :)

I think it might be a good idea to warn the Jore4 user if the same stop point is used more than once by a journey pattern, though. But maybe that's a different story and need.
</blockquote></details>

The new version supports all possible kinds of dead-ends and loops and actually validates the journey pattern by traversing it. I also added an additional integration test, which inserts the (most crucial parts) of the route and journey pattern mentioned by @datarttu .

jarkkoka (approved):
Props for clever usage of recursive CTE :lgtm: 

I re-reviewed the SQL part. Did not spot errors by eye. Did only lightly view the TS test code. Now that there are more proficient UI guys in the lineup they might be more relevant people to review that part of the PR.

> Reviewed 11 of 55 files at r1, 5 of 7 files at r3, 1 of 1 files at r5, 11 of 11 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-:-MuGPcTB4d-Josi0te5B:b-vvil3o)* status: all files reviewed, 7 unresolved discussions (waiting on @jarkkoka and @jschummer)

___
*[integration-test/datasets/route116/scheduled-stop-points.ts, line 23 at r6](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MuG_8Pp0bISNs3i2B-N:-MuG_8Pp0bISNs3i2B-O:b-3n1ki0) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/722679bbdd49e25ebd495172b44f78d6eae78d31/integration-test/datasets/route116/scheduled-stop-points.ts#L23)):*
> ```XML
>     label: "WeeGee-talo",
>     priority: 10,
>     validity_start: null,
> ```

Pretty much redundant data but never mind (could have been inserted in a loop).



_Code quote:_
```XML
    priority: 10,
    validity_start: null,
    validity_end: null,
```
___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts, line 55 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNqluB9_0ufXF0lfRU:-MuG_NsIBxeqb4s51H8R:b-df54rn) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/delete-infra-link-from-route.spec.ts#L55)):*

Yes, I do see your point, I guess it was a bit misleading. Maybe the new form ("must match") is clearer?
</blockquote></details>

It's good now. :+1: 
___
*[integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts, line 122 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MtNt7ku8gxUsx_WMNl3:-MuGalrG6hk9dBRbzcrl:b1idazm) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/7b82dc94f0a8c992f0614891e0d95a5558e1ecd4/integration-test/tests/data-integrity/route-journey-pattern-refs/insert-stop-into-journey-pattern.spec.ts#L122)):*

I now changed the occurrance(s) of "compatible" to "matching" in this PR. There are still some left from previous PRs. If these should be changed too, that would require follow-up work.
</blockquote></details>

Ok. No worries. Let's keep it what it is now and improve later...
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 43 at r6](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MuG0w_d8P9KNyeImxnV:-MuG0w_d8P9KNyeImxnW:bltbard) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/722679bbdd49e25ebd495172b44f78d6eae78d31/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L43)):*
> ```TSQL
>                          OVER (PARTITION BY journey_pattern_id ORDER BY scheduled_stop_point_sequence) AS stop_point_order
>                   FROM journey_pattern.scheduled_stop_point_in_journey_pattern
>                 ) AS sspijp
> ```

Awkward, unfriendly acronym. E.g. `stop_info` or `stop_on_pattern` would be much more readable/understandable.



_Code quote:_
```TSQL
                ) AS sspijp
```
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 56 at r6](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MuGBIpq1fy-Z9SyaqV8:-MuGBIpq1fy-Z9SyaqV9:bdsnim) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/722679bbdd49e25ebd495172b44f78d6eae78d31/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L56)):*
> ```TSQL
>                                      OR (ssp.direction = 'backward' AND ilar.is_traversal_forwards = false)))
>            WHERE jp.journey_pattern_id = filter_journey_pattern_id
>               OR jp.on_route_id = filter_route_id
> ```

It is not enforced that `filter_journey_pattern_id` and `filter_route_id` are actually related to each other. Wondering whether this could yield incorrect results in case both are given and they are not related??...



_Code quote:_
```TSQL
           WHERE jp.journey_pattern_id = filter_journey_pattern_id
              OR jp.on_route_id = filter_route_id
```
___
*[migrations/default/1641064819374_route_start_end_stop_point_link_constraints/up.sql, line 2 at r6](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MuGROSg4a2L3Nn1qyuP:-MuGROSg4a2L3Nn1qyuQ:btyr1k6) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/722679bbdd49e25ebd495172b44f78d6eae78d31/migrations/default/1641064819374_route_start_end_stop_point_link_constraints/up.sql#L2)):*
> ```TSQL
> 
> CREATE FUNCTION route.verify_route_start_end_stop_points(param_route_id UUID)
> ```

I must admit that I'm not sure whether this constraint corresponds to the original intent of the stop points defined in route overview.

However, those stop points, when required to locate on terminus links, can be used to crop the geometric presentation of route.



_Code quote:_
```TSQL
route.verify_route_start_end_stop_points(param_route_id UUID)
```

jschummer (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42)* status: all files reviewed, 7 unresolved discussions (waiting on @jarkkoka and @jschummer)

___
*[integration-test/datasets/route116/scheduled-stop-points.ts, line 23 at r6](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MuG_8Pp0bISNs3i2B-N:-MuKGaxp6EgbKig2gJsz:b-dbz7ml) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/722679bbdd49e25ebd495172b44f78d6eae78d31/integration-test/datasets/route116/scheduled-stop-points.ts#L23)):*

Pretty much redundant data but never mind (could have been inserted in a loop).
</blockquote></details>

Yes, true, but it's only the test data, so I won't bother cleaning this up.
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 43 at r6](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MuG0w_d8P9KNyeImxnV:-MuKI2OGEjBrTva_Hcv9:bpesrdm) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/722679bbdd49e25ebd495172b44f78d6eae78d31/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L43)):*

Awkward, unfriendly acronym. E.g. `stop_info` or `stop_on_pattern` would be much more readable/understandable.
</blockquote></details>

Well, I'm not sure either one you suggested couldn't be misinterpreted in some way, so I'd rather stick with this one (which my SQL client suggested to me :-) ).
___
*[migrations/default/1640981605602_route_journey_pattern_constraints/up.sql, line 56 at r6](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MuGBIpq1fy-Z9SyaqV8:-MuKIFqL-IFAxPqlQTbj:bbop9iy) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/722679bbdd49e25ebd495172b44f78d6eae78d31/migrations/default/1640981605602_route_journey_pattern_constraints/up.sql#L56)):*

It is not enforced that `filter_journey_pattern_id` and `filter_route_id` are actually related to each other. Wondering whether this could yield incorrect results in case both are given and they are not related??...
</blockquote></details>

Yes, they could be unrelated. But in that case the function would simply check
- the journey pattern with the given filter_journey_pattern_id is intact and
- all journey patterns of the route with the given filter_route_id are intact.

I can't say what the use case for this would be, but I think this would be logical behaviour and shouldn't break anything.
___
*[migrations/default/1641064819374_route_start_end_stop_point_link_constraints/up.sql, line 2 at r6](https://reviewable.io/reviews/hsldevcom/jore4-hasura/42#-MuGROSg4a2L3Nn1qyuP:-MuKH4g8-SZyMG9y0v2a:b-g4g9sl) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/722679bbdd49e25ebd495172b44f78d6eae78d31/migrations/default/1641064819374_route_start_end_stop_point_link_constraints/up.sql#L2)):*

I must admit that I'm not sure whether this constraint corresponds to the original intent of the stop points defined in route overview.

However, those stop points, when required to locate on terminus links, can be used to crop the geometric presentation of route.
</blockquote></details>

I don't quite understand what you mean. I agree, that the (main) purpose of the route start and end stop points is the cropping of geometry (of course I'd assume the route really starts / ends at those points in that case).

(As you noticed, this constraint does not affect the journey pattern, so this info shouldn't be used to display passenger relevant information.)

PR #83, opened 2022-06-30
Allow multiple overlapping draft versions

culka:
Allow overlapping draft versions of stop points, routes and lines to exist in the database.

---
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/83)

### DIFF
--- a/integration-test/datasets/defaultSetup/routes.ts
+++ b/integration-test/datasets/defaultSetup/routes.ts
@@ -45,6 +45,15 @@ export const routes: Route[] = [
     validity_start: new Date('2044-01-02 21:11:32Z'),
     validity_end: new Date('2044-09-02 22:11:32Z'),
   },
+  {
+    route_id: '7dac4416-1f84-4951-86b7-169d00bcd8d9',
+    on_line_id: lines[4].line_id,
+    ...buildRoute('5'),
+    direction: RouteDirection.Eastbound,
+    priority: 20,
+    validity_start: null,
+    validity_end: new Date('2044-09-02 23:11:32Z'),
+  },
 ];
 
 export const infrastructureLinkAlongRoute: InfrastructureLinkAlongRoute[] = [
--- a/integration-test/datasets/defaultSetup/scheduled-stop-points.ts
+++ b/integration-test/datasets/defaultSetup/scheduled-stop-points.ts
@@ -59,7 +59,7 @@ export const scheduledStopPoints: ScheduledStopPoint[] = [
       },
     },
     label: 'stop2',
-    priority: 30,
+    priority: 20,
     validity_start: null,
     validity_end: new Date('2064-01-03 12:34:56'),
   },
--- a/integration-test/tests/priority-and-validity-period-constraints/line/insert-without-conflicts.spec.ts
+++ b/integration-test/tests/priority-and-validity-period-constraints/line/insert-without-conflicts.spec.ts
@@ -87,7 +87,7 @@ describe('Insert line', () => {
       ...buildLine('34', VehicleMode.Tram),
       name_i18n: buildLocalizedString('conflicting transport tram line 34'),
       short_name_i18n: buildLocalizedString('conflicting line 34'),
-      priority: 30,
+      priority: 10,
       validity_start: new Date('2041-06-01 23:11:32Z'),
       validity_end: new Date('2042-06-01 23:11:32Z'),
     };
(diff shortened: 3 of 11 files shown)


jschummer (changes requested):
Good work!

Do you think the second commit ("Correct internal_route") could be squashed into the first one? I think there is no value to have it on its own, because it's rather a fixup if I understand correctly. The tests can be separate IMO :+1: 

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/83#-:-N679MkkE6Xc_WyCCarD:bc0ncf0)* status: 0 of 9 files reviewed, 2 unresolved discussions (waiting on @culka and @jschummer)

___
*[`integration-test/tests/priority-and-validity-period-constraints/route/insert-without-conflicts.spec.ts` line 162 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/83#-N6799yk2vUm-_aFCxQt:-N6799yk2vUm-_aFCxQu:b-ic86xe) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/18a5a34c83a6dfb2ea88c34d3d6e37177380ff8e/integration-test/tests/priority-and-validity-period-constraints/route/insert-without-conflicts.spec.ts#L162)):*
> ```XML
>   });
> 
>   describe('whose validity period overlaps with existing validity but both have draft priority', () => {
> ```

Nice you added tests for this!



_Code quote:_
```XML
 describe('whose validity period overlaps with existing validity but both have draft priority',
```
___
*[`migrations/generic/default/1656589426929_allow_multiple_overlapping_drafts/up.sql` line 12 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/83#-N672cID5CRxmsOTiQ5q:-N672cID5CRxmsOTiQ5r:b-1bwb3f) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/18a5a34c83a6dfb2ea88c34d3d6e37177380ff8e/migrations/generic/default/1656589426929_allow_multiple_overlapping_drafts/up.sql#L12)):*
> ```TSQL
>   TSTZRANGE(validity_start, validity_end) WITH &&
> )
> WHERE (priority < 30);
> ```

Now that this constant is getting a semantic meaning inside our SQL code, I'm wondering if we can define this somehow globally.

Apparently there's no way of defining a real constant in Postgresql, but I came across a workaround of defining a global function: https://stackoverflow.com/questions/13316773/is-there-a-way-to-define-a-named-constant-in-a-postgresql-query

Do you think something like this would be feasible here?



_Code quote:_
```TSQL
30
```

culka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/83)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @culka and @jschummer)

___
*[`migrations/generic/default/1656589426929_allow_multiple_overlapping_drafts/up.sql` line 12 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/83#-N672cID5CRxmsOTiQ5q:-N6CFJx9AjiWp2PfvIrg:b30d1e1) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/18a5a34c83a6dfb2ea88c34d3d6e37177380ff8e/migrations/generic/default/1656589426929_allow_multiple_overlapping_drafts/up.sql#L12)):*

Now that this constant is getting a semantic meaning inside our SQL code, I'm wondering if we can define this somehow globally.

Apparently there's no way of defining a real constant in Postgresql, but I came across a workaround of defining a global function: https://stackoverflow.com/questions/13316773/is-there-a-way-to-define-a-named-constant-in-a-postgresql-query

Do you think something like this would be feasible here?
</blockquote></details>

Added internal_utils.const_priority -functions for constant priority values.

culka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/83)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @culka and @jschummer)

___
*[`migrations/generic/default/1656589426778_constant_values_for_priorities/up.sql` line 6 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/83#-N6gCAaQ8zrM_mqPK1rJ:-N6l4ARf2Eb-GQzwKne7:b-ticrb4) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/fbb52117c111231ab5250173cf36305584b43897/migrations/generic/default/1656589426778_constant_values_for_priorities/up.sql#L6)):*

Are these first two used somewhere? If not, I'd rather leave them out, because there is the danger that they'll get outdated. And I hope we don't need to know these other values on this side of the data model... :crossed_fingers: 

But a very good solution to define those constants :+1: 
</blockquote></details>

Removed them!

jschummer (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/83#-:-N6rMmza6fvjWkCghcHF:bnfp4nl)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @culka and @jschummer)

PR #96, opened 2022-09-27
Beta: Initial version of the timetables data model

pallost:
this is a work-in-progress version of the timetables data model. The intension is to have the migration in a reviewable format while also providing a usable version for UI development

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)

Closes https://github.com/HSLdevcom/jore4/issues/681


pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: 0 of 35 files reviewed, 1 unresolved discussion

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 95 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRXyYX0aDjJUfPea6O:-NDRXyYX0aDjJUfPea6P:b-wq73a6) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/ed43ac77559b9abedd36e5b5389fda7841f5a5a6/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L95)):*
> ```TSQL
> COMMENT ON COLUMN vehicle_service.vehicle_service.vehicle_schedule_frame_id IS 'Human-readable name for the VEHICLE SCHEDULE FRAME';
> 
> CREATE TABLE vehicle_service.block (
> ```

Transmodel defines multiple layers under VEHICLE SERVICE. I still don't see however what their meaning actually would be in HSL context.

"The whole work of a logical vehicle for a DAY TYPE is described by a VEHICLE SERVICE, composed of one or several VEHICLE SERVICE PARTs. Three levels of activities (BLOCK, VEHICLE SERVICE PART, and VEHICLE SERVICE) are possible, but will not always be described. The concept of VEHICLE SERVICE is not strictly necessary, once the number of required vehicles has been optimised and, in many situations, the distinction between BLOCK and VEHICLE SERVICE PART is not useful. Therefore quite often only a subpart of this model will be used (e.g. BLOCK)." (page 184 from Public transport. Network and Timetable Exchange (NeTEx). Part 2 Public transport scheduled timetables exchange format.pdf)

1. I think for us either the BLOCK or the VEHICLE SERVICE entity should be enough, there's no need to split a single logical vehicle's full-day journey into multiple BLOCKs.

2. I've been wondering that where should we store data about the requirements for the vehicle (should it be a big bus, small bus, etc). To my understanding, these are marked on the VEHICLE JOURNEY level in JORE3. However I think it would be more logical to put this information to VEHICLE SERVICE in our data model

jarkkoka (commented):
> Reviewed 11 of 33 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: 11 of 35 files reviewed, 5 unresolved discussions (waiting on @jarkkoka and @pallost)

___
*[`migrations/generic/default/1663669245583_timetables_fixes/up.sql` line 13 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRWeI_BAB-y4k7AEqH:-NDRWeI_BAB-y4k7AEqI:b796nkx) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/default/1663669245583_timetables_fixes/up.sql#L13)):*
> ```TSQL
>   UNIQUE (label)
> );
> COMMENT ON TABLE timing_pattern.timing_point IS 'A POINT against which the timing information necessary to build schedules may be recorded. In HSL context, this is "Hastus paikka". Transmodel: https://www.transmodel-cen.eu/model/index.htm?goto=2:3:2:709 ';
> ```

I don't think this is correct. In Transmodel, timing point is such a point in journey pattern (Note! only scheduled stop points in HSL context) for which timing information is necessary to build schedules.

I think, in HSL context, "Hastus-paikka" is a static property of scheduled stop point and not dependent on to which journey pattern it is included. Hastus-paikka groups multiple scheduled stop points under a common label.



_Code quote:_
```TSQL
In HSL context, this is "Hastus paikka"
```
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 9 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRh93w0uc14GG8ibvS:-NDRh93w0uc14GG8ibvT:b4dn4k3) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L9)):*
> ```TSQL
>   journey_pattern_ref_id UUID DEFAULT gen_random_uuid() PRIMARY KEY,
>   journey_pattern_id UUID NOT NULL,
>   observation_date_timestamp TIMESTAMP WITH TIME ZONE NOT NULL,
> ```

`date_timestamp` postfix includes two different temporal types (`date` and `timestamp`) which is confusing from a DB design standpoint. Please choose only one. I suggest `observation_timestamp`



_Code quote:_
```TSQL
observation_date_timestamp
```
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 14 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRgbDl8VVbL-OCIQ2a:-NDRgbDl8VVbL-OCIQ2b:b-tnfu87) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L14)):*
> ```TSQL
> COMMENT ON TABLE journey_pattern.journey_pattern_ref IS 'Reference to a given snapshot of a JOURNEY PATTERN for a given operating day. Transmodel: https://www.transmodel-cen.eu/model/index.htm?goto=2:3:4:729 ';
> COMMENT ON COLUMN journey_pattern.journey_pattern_ref.journey_pattern_id IS 'The ID of the referenced JOURNEY PATTERN';
> COMMENT ON COLUMN journey_pattern.journey_pattern_ref.observation_date_timestamp IS 'The observation date of which the JOURNEY PATTERN snapshot was taken for';
> ```

My suggestion: "The user-given point of time used to pick one journey pattern (route) among possibly many variants. The selected, unambiguous journey pattern variant is used as a basis for schedule planning."



_Code quote:_
```TSQL
The observation date of which the JOURNEY PATTERN snapshot was taken for
```
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 15 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRZW3_B5zQyMc4ybdy:-NDRZW3_B5zQyMc4ybdz:be0tlqk) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L15)):*
> ```TSQL
> COMMENT ON COLUMN journey_pattern.journey_pattern_ref.journey_pattern_id IS 'The ID of the referenced JOURNEY PATTERN';
> COMMENT ON COLUMN journey_pattern.journey_pattern_ref.observation_date_timestamp IS 'The observation date of which the JOURNEY PATTERN snapshot was taken for';
> COMMENT ON COLUMN journey_pattern.journey_pattern_ref.snapshot_timestamp IS 'The date when the snapshot was taken';
> ```

These sound like a same thing: "a timestamp for a snapshot"



_Code quote:_
```TSQL
COMMENT ON COLUMN journey_pattern.journey_pattern_ref.observation_date_timestamp IS 'The observation date of which the JOURNEY PATTERN snapshot was taken for';
COMMENT ON COLUMN journey_pattern.journey_pattern_ref.snapshot_timestamp IS 'The date when the snapshot was taken';
```

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: 11 of 35 files reviewed, 4 unresolved discussions (waiting on @jarkkoka)

___
*[`migrations/generic/default/1663669245583_timetables_fixes/up.sql` line 13 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRWeI_BAB-y4k7AEqH:-NDSUOxT35IBYcsdQryw:b-wfixpu) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/default/1663669245583_timetables_fixes/up.sql#L13)):*

I don't think this is correct. In Transmodel, timing point is such a point in journey pattern (Note! only scheduled stop points in HSL context) for which timing information is necessary to build schedules.

I think, in HSL context, "Hastus-paikka" is a static property of scheduled stop point and not dependent on to which journey pattern it is included. Hastus-paikka groups multiple scheduled stop points under a common label.
</blockquote></details>

I do agree, this is not completely correct. Even though semantically these are TIMING POINTs, we don't use them the same way as Transmodel defines them. Would it be clearer if we used a different name? E.g. HASTUS PLACE?
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 9 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRh93w0uc14GG8ibvS:-NDSUp8XByJEe6MiBlIZ:b-hjz5ka) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L9)):*

`date_timestamp` postfix includes two different temporal types (`date` and `timestamp`) which is confusing from a DB design standpoint. Please choose only one. I suggest `observation_timestamp`
</blockquote></details>

good point, fixing it.
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 14 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRgbDl8VVbL-OCIQ2a:-NDSVc0PBHuneN0b56iw:brzd1gq) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L14)):*

My suggestion: "The user-given point of time used to pick one journey pattern (route) among possibly many variants. The selected, unambiguous journey pattern variant is used as a basis for schedule planning."
</blockquote></details>

thank you for the description. I extended it with clarification about scheduled stop points too
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 15 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRZW3_B5zQyMc4ybdy:-NDSVlOa9q3gJqkYCaLm:b-mou9yb) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L15)):*

These sound like a same thing: "a timestamp for a snapshot"
</blockquote></details>

Would this be better like this?

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: 11 of 35 files reviewed, 4 unresolved discussions (waiting on @jarkkoka)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 15 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRZW3_B5zQyMc4ybdy:-NDSW__z4ryziMzmIAgT:b1k38to) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L15)):*

Would this be better like this?
</blockquote></details>

This comment is obsolete once I figured out the meaning of the other timestamp field. I posted this comment accidentally. Tried to retract it...

jschummer (changes requested):
Thank you very much for a good start into the concrete modelling process. Like mentioned elsewhere, I too have to think this one through a couple of more times before being able to give more comments...

> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-:-NDUKiAx6_UU5MGciRKC:bi27c29)* status: 11 of 35 files reviewed, 7 unresolved discussions (waiting on @jarkkoka, @jschummer, and @pallost)

___
*[`migrations/generic/default/1663669245583_timetables_fixes/up.sql` line 13 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRWeI_BAB-y4k7AEqH:-NDUB9Kc5EPM0tlJpVhe:bb0hq6o) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/default/1663669245583_timetables_fixes/up.sql#L13)):*

I do agree, this is not completely correct. Even though semantically these are TIMING POINTs, we don't use them the same way as Transmodel defines them. Would it be clearer if we used a different name? E.g. HASTUS PLACE?
</blockquote></details>

I think (...) what Jarkko meant to say was that it should / could say something like

`In HSL context, those scheduled stop points are modeled as timing points, which are declared to be (part of) a so-called Hastus-place.`
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 95 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRXyYX0aDjJUfPea6O:-NDUDZ-M5zOZ9UG9GJl9:bxsjvw3) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/ed43ac77559b9abedd36e5b5389fda7841f5a5a6/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L95)):*

Transmodel defines multiple layers under VEHICLE SERVICE. I still don't see however what their meaning actually would be in HSL context.

"The whole work of a logical vehicle for a DAY TYPE is described by a VEHICLE SERVICE, composed of one or several VEHICLE SERVICE PARTs. Three levels of activities (BLOCK, VEHICLE SERVICE PART, and VEHICLE SERVICE) are possible, but will not always be described. The concept of VEHICLE SERVICE is not strictly necessary, once the number of required vehicles has been optimised and, in many situations, the distinction between BLOCK and VEHICLE SERVICE PART is not useful. Therefore quite often only a subpart of this model will be used (e.g. BLOCK)." (page 184 from Public transport. Network and Timetable Exchange (NeTEx). Part 2 Public transport scheduled timetables exchange format.pdf)

1. I think for us either the BLOCK or the VEHICLE SERVICE entity should be enough, there's no need to split a single logical vehicle's full-day journey into multiple BLOCKs.

2. I've been wondering that where should we store data about the requirements for the vehicle (should it be a big bus, small bus, etc). To my understanding, these are marked on the VEHICLE JOURNEY level in JORE3. However I think it would be more logical to put this information to VEHICLE SERVICE in our data model
</blockquote></details>

1. I agree that we most probably don't need to model all 3 possible layers. Not sure myself (yet) which ones to pick though. I think BLOCKs need to be modelled at some point, because they are the bases for compensation calculations, if I understood correctly. Also Hastus planning is based on blocks. But for the timetables they are not strictly needed and maybe we don't need to model them now (unless we can combine both Hastus imports - containing timetables and compensation information - into one).

2. I think the exact vehicle type probably does not appear anywhere in connection with the timetables. But we maybe need to make sure if some sub-specification (e.g. "low floor bus") is used on the timetable side. Otherwise this information might be better kept with compensation information. 
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 42 at r3](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDUIpHZ1bwYwZg27APh:-NDUIpHZ1bwYwZg27APi:b47h68w) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/22af55f2f669c71abe8b199a95c52ef243f6f7a2/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L42)):*
> ```TSQL
>   day_type_id UUID DEFAULT gen_random_uuid() PRIMARY KEY,
>   label TEXT NOT NULL,
>   name JSONB NOT NULL,
> ```

Should or could this be `name_i18n`?



_Code quote:_
```TSQL
name
```
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 43 at r3](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDUJ4khC_4CgszwaYZA:-NDUJ4khC_4CgszwaYZB:b-qg1upu) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/22af55f2f669c71abe8b199a95c52ef243f6f7a2/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L43)):*
> ```TSQL
>   label TEXT NOT NULL,
>   name JSONB NOT NULL,
>   priority int NOT NULL,
> ```

I'm not totally sure if we should model priorities at this level or rather only together with the mapping of day types to calendar dates (`vehicle_schedule_frame` or `vehicle_service`?). Have to think this one through more thoroughly.



_Code quote:_
```TSQL
priority int NOT NULL,
```
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 70 at r3](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDUJQzQ1R3zyvZmF_jL:-NDUJQzQ1R3zyvZmF_jM:b-eroj2v) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/22af55f2f669c71abe8b199a95c52ef243f6f7a2/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L70)):*
> ```TSQL
>   vehicle_schedule_frame_id UUID DEFAULT gen_random_uuid() PRIMARY KEY,
>   label TEXT NOT NULL,
>   name JSONB NOT NULL,
> ```

Same as above, `name_i18n`?



_Code quote:_
```TSQL
name
```

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: 11 of 35 files reviewed, 8 unresolved discussions (waiting on @jarkkoka, @jschummer, and @pallost)

___
*[`migrations/generic/default/1663669245583_timetables_fixes/up.sql` line 13 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRWeI_BAB-y4k7AEqH:-NDXPsGf5g9ijEeuym55:b-4fh5vz) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/default/1663669245583_timetables_fixes/up.sql#L13)):*

I think (...) what Jarkko meant to say was that it should / could say something like

`In HSL context, those scheduled stop points are modeled as timing points, which are declared to be (part of) a so-called Hastus-place.`
</blockquote></details>

Jörg, do you really think I accidentally said something else what I meant? Not in this case. I stand by my words above. Actually, what you claim that was my intention is not correct since I do not fully understand your quote. What do you refer by saying `those scheduled stop points`? Which scheduled stop points? I think what you said is your own opinion, not mine. Honestly, to me what you quoted can be understood like

```
In Jore4, all scheduled stop points are modeled as timing points.
```

And that does not make sense to me since `is_timing_point` is a property of `scheduled_stop_point_in_journey_pattern` and hence `scheduled_stop_point != timing_point` because whether a scheduled stop point is also a timing point depends on which journey patter the stop point is part of.

Of course things can be said in many ways and some expressions may be more clear to someone else depending on the listener. Communication is not easy. :)
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 57 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDSnHrL2POB5eVMQl3w:-NDSnHrL2POB5eVMQl3x:b-nzqobp) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L57)):*
> ```TSQL
>   ('6781bd06-08cf-489e-a2bb-be9a07b15752', 'MT','{"fi_FI": "Maanantai - Torstai", "sv_FI": "Måndag - Torsdag"}', 10),
>   ('38853b0d-ec36-4110-b4bc-f53218c6cdcd', 'MP','{"fi_FI": "Maanantai - Perjantai", "sv_FI": "Måndag - Fredag"}', 10),
>   ('7176e238-d46e-4583-a567-b836b1ae2589', 'PE','{"fi_FI": "Perjantai", "sv_FI": "Fredag"}', 10),
> ```

Shouldn't this priority be higher than what is defined for "Maanantai - Perjantai"? Currently, it is undefined what happens if both were present.

Or is it error condition if both "Maanantai - Perjantai" and "Perjantai" are present?



_Code quote:_
```TSQL
10
```
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 64 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDSo6Vb9voML2Uqtnjo:-NDSo6Vb9voML2Uqtnjp:b-s56lji) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L64)):*
> ```TSQL
> 
> CREATE SCHEMA vehicle_schedule;
> COMMENT ON SCHEMA vehicle_schedule IS 'The vehicle schedule frame adapted from Transmodel: https://www.transmodel-cen.eu/model/index.htm?goto=3:7:2:993 ';
> ```

Minor detail: There is an extra whitespace character after Transmodel hyperlink. This repeats also in many other comments with trailing hyperlink.
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 132 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDTY6n46HXmwG_3TmMm:-NDTY6n5-Z3rB3LC_FGJ:b-lu5ezz) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L132)):*
> ```TSQL
> COMMENT ON COLUMN passing_times.timetabled_passing_time.scheduled_stop_point_in_journey_pattern_ref_id IS 'The SCHEDULED STOP POINT of the JOURNEY PATTERN where the vehicle passes';
> COMMENT ON COLUMN passing_times.timetabled_passing_time.arrival_time IS 'The time when the vehicle arrives to the SCHEDULED STOP POINT. Measured as interval counted from the midnight of the OPERATING DAY';
> COMMENT ON COLUMN passing_times.timetabled_passing_time.departure_time IS 'The time when the vehicle departs from the SCHEDULED STOP POINT. Measured as interval counted from the midnight of the OPERATING DAY';
> ```

When null then what? I would prefer to have this mentioned in the comment.



_Code quote:_
```TSQL
The time when the vehicle departs from the SCHEDULED STOP POINT. Measured as interval counted from the midnight of the OPERATING DAY
```

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: 11 of 35 files reviewed, 8 unresolved discussions (waiting on @jarkkoka, @jschummer, and @pallost)

___
*[`migrations/generic/default/1663669245583_timetables_fixes/up.sql` line 13 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRWeI_BAB-y4k7AEqH:-NDXwSztE5co3_AcSghR:b-gilrfd) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/default/1663669245583_timetables_fixes/up.sql#L13)):*

Jörg, do you really think I accidentally said something else what I meant? Not in this case. I stand by my words above. Actually, what you claim that was my intention is not correct since I do not fully understand your quote. What do you refer by saying `those scheduled stop points`? Which scheduled stop points? I think what you said is your own opinion, not mine. Honestly, to me what you quoted can be understood like

```
In Jore4, all scheduled stop points are modeled as timing points.
```

And that does not make sense to me since `is_timing_point` is a property of `scheduled_stop_point_in_journey_pattern` and hence `scheduled_stop_point != timing_point` because whether a scheduled stop point is also a timing point depends on which journey patter the stop point is part of.

Of course things can be said in many ways and some expressions may be more clear to someone else depending on the listener. Communication is not easy. :)
</blockquote></details>

Ok. Re-read the whole conversation. Now I understood that quite likely Jörg did not mean to correct what I was saying. But to replace the sentence in the comment starting with "In HSL context,". So, maybe there was a misinterpretation from my part, sorry.

However, the first sentence in the comment ("A POINT is ...") should also be changed to reflect the meaning in HSL context since there are possibly many points (scheduled stop points) that are associated with a single Hastus place.

Overall, I'm uncertain about how these Hastus places can be cleanly associated with scheduled stop point records (with data integrity) that reside in different logical database. IMO it would not be a big mistake to introduce Hastus places in the network & lines database since most data in Jore->Hastus exports originates from that database. Do these Hastus places appear also in Hastus->Jore imports? And are these Hastus places used to map back to actual scheduled stop points?



_Code quote:_
```TSQL
In HSL context, this is "Hastus paikka".
```

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: 11 of 35 files reviewed, 8 unresolved discussions (waiting on @jarkkoka, @jschummer, and @pallost)

___
*[`migrations/generic/default/1663669245583_timetables_fixes/up.sql` line 13 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRWeI_BAB-y4k7AEqH:-NDgzD1V9ck9sG1DD4d9:b-onf7kh) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/default/1663669245583_timetables_fixes/up.sql#L13)):*

Ok. Re-read the whole conversation. Now I understood that quite likely Jörg did not mean to correct what I was saying. But to replace the sentence in the comment starting with "In HSL context,". So, maybe there was a misinterpretation from my part, sorry.

However, the first sentence in the comment ("A POINT is ...") should also be changed to reflect the meaning in HSL context since there are possibly many points (scheduled stop points) that are associated with a single Hastus place.

Overall, I'm uncertain about how these Hastus places can be cleanly associated with scheduled stop point records (with data integrity) that reside in different logical database. IMO it would not be a big mistake to introduce Hastus places in the network & lines database since most data in Jore->Hastus exports originates from that database. Do these Hastus places appear also in Hastus->Jore imports? And are these Hastus places used to map back to actual scheduled stop points?
</blockquote></details>

Sorry, didn't previously notice that this `timing_pattern.timing_point` table was already defined in the network & lines db. Turned out in Slack conversation.

I think I can approve this table definition along with the comment as it is now that we have discussed about the issue in Slack in more detail.

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: 9 of 35 files reviewed, 8 unresolved discussions (waiting on @jarkkoka and @jschummer)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 57 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDSnHrL2POB5eVMQl3w:-NE4cffr4wNqp7x1gHOj:bnkzfbs) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L57)):*

Shouldn't this priority be higher than what is defined for "Maanantai - Perjantai"? Currently, it is undefined what happens if both were present.

Or is it error condition if both "Maanantai - Perjantai" and "Perjantai" are present?
</blockquote></details>

Removed priorities from day types
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 64 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDSo6Vb9voML2Uqtnjo:-NDb524dAyEg3z0fcxQQ:bljjha3) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L64)):*

Minor detail: There is an extra whitespace character after Transmodel hyperlink. This repeats also in many other comments with trailing hyperlink.
</blockquote></details>

This is intentional, so that the ending `'` doesn't become part of the url when trying to open the link from a code editor. It's much faster to navigate like this.
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 132 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDTY6n46HXmwG_3TmMm:-NE4oWEW4_BN82FVnQv5:b7ib5xr) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L132)):*

When null then what? I would prefer to have this mentioned in the comment.
</blockquote></details>

good point. Added it
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 95 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRXyYX0aDjJUfPea6O:-NE4ix0-81L3QlW5Gprb:b-x44l95) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/ed43ac77559b9abedd36e5b5389fda7841f5a5a6/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L95)):*

1. I agree that we most probably don't need to model all 3 possible layers. Not sure myself (yet) which ones to pick though. I think BLOCKs need to be modelled at some point, because they are the bases for compensation calculations, if I understood correctly. Also Hastus planning is based on blocks. But for the timetables they are not strictly needed and maybe we don't need to model them now (unless we can combine both Hastus imports - containing timetables and compensation information - into one).

2. I think the exact vehicle type probably does not appear anywhere in connection with the timetables. But we maybe need to make sure if some sub-specification (e.g. "low floor bus") is used on the timetable side. Otherwise this information might be better kept with compensation information. 
</blockquote></details>

Let's return to the vehicle types later. Blocks should however be taken into use, they are also modeled in Hastus and could come handy also later on
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 42 at r3](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDUIpHZ1bwYwZg27APh:-NE4cYXTEWoy-EU-AN_o:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/22af55f2f669c71abe8b199a95c52ef243f6f7a2/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L42)):*

Should or could this be `name_i18n`?
</blockquote></details>

Done.
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 43 at r3](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDUJ4khC_4CgszwaYZA:-NE4cZkb8jdr6TYdxxo1:buka82n) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/22af55f2f669c71abe8b199a95c52ef243f6f7a2/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L43)):*

I'm not totally sure if we should model priorities at this level or rather only together with the mapping of day types to calendar dates (`vehicle_schedule_frame` or `vehicle_service`?). Have to think this one through more thoroughly.
</blockquote></details>

Removed it for now.
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 70 at r3](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDUJQzQ1R3zyvZmF_jL:-NE4cl1r4XlcitCAdndt:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/22af55f2f669c71abe8b199a95c52ef243f6f7a2/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L70)):*

Same as above, `name_i18n`?
</blockquote></details>

Done.
___
*[`migrations/generic/default/1663669245583_timetables_fixes/up.sql` line 13 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRWeI_BAB-y4k7AEqH:-NE-pqZY2kASxhn49Z2m:bns81mw) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/6316cc23efa09c421f40425a695f5aa9bc3979b1/migrations/generic/default/1663669245583_timetables_fixes/up.sql#L13)):*

Sorry, didn't previously notice that this `timing_pattern.timing_point` table was already defined in the network & lines db. Turned out in Slack conversation.

I think I can approve this table definition along with the comment as it is now that we have discussed about the issue in Slack in more detail.
</blockquote></details>

Thank you for your comments, they started a great conversation in Slack about how Hastus works. Done.

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: 9 of 36 files reviewed, 9 unresolved discussions (waiting on @jarkkoka and @jschummer)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 71 at r6](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NE5409g3QmoqsiJusQw:-NE5409g3QmoqsiJusQx:b-s0bzlp) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/102a720551dc10c73d30fab3106514a9eee79fbe/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L71)):*
> ```TSQL
>   validity_start DATE NULL,
>   validity_end DATE NULL,
>   UNIQUE (label)
> ```

it came to my mind that should the label actually really be unique. Or that do we need a label at all? For timetables, priorities are calculated on an OPERATING DAY basis, not based on labels.

IMHO labels are only just "short name" fields in our case, they are not used to identify anything.

jschummer (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: 9 of 36 files reviewed, 8 unresolved discussions (waiting on @jarkkoka and @pallost)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 95 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRXyYX0aDjJUfPea6O:-NEA5M4ODf_wmSCetHVo:b-g70mvb) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/ed43ac77559b9abedd36e5b5389fda7841f5a5a6/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L95)):*

Let's return to the vehicle types later. Blocks should however be taken into use, they are also modeled in Hastus and could come handy also later on
</blockquote></details>

I think they definitely need to be modelled when we import data from Hastus. But I agree that this is not necessarily something that needs to be done in this PR.
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 71 at r6](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NE5409g3QmoqsiJusQw:-NEA592i-BWSbvzvDYVC:b-etlxs7) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/102a720551dc10c73d30fab3106514a9eee79fbe/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L71)):*

it came to my mind that should the label actually really be unique. Or that do we need a label at all? For timetables, priorities are calculated on an OPERATING DAY basis, not based on labels.

IMHO labels are only just "short name" fields in our case, they are not used to identify anything.
</blockquote></details>

I'm not sure either what this would be used for. If there is no clear use case, I think it's better to remove it.

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: 7 of 39 files reviewed, 8 unresolved discussions (waiting on @jarkkoka and @jschummer)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 71 at r6](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NE5409g3QmoqsiJusQw:-NF8AwHE3VrhIIOI-A4Q:brczp7e) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/102a720551dc10c73d30fab3106514a9eee79fbe/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L71)):*

I'm not sure either what this would be used for. If there is no clear use case, I think it's better to remove it.
</blockquote></details>

removed. Can readd later if we find the need

jarkkoka (changes requested):
> Reviewed 12 of 33 files at r1, 2 of 2 files at r2, 9 of 15 files at r8, 6 of 6 files at r9, 1 of 3 files at r10, 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: all files reviewed, 8 unresolved discussions (waiting on @jschummer and @pallost)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 9 at r11](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NFEZvC-DKzVH2dt0BT6:-NFEZvC09rv68MUmIGu0:bvmqzoc) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a8e16aa0a21cc91747c19d97bff5d2708c29c825/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L9)):*
> ```TSQL
>   journey_pattern_ref_id UUID DEFAULT gen_random_uuid() PRIMARY KEY,
>   journey_pattern_id UUID NOT NULL,
>   observation_timestamp TIMESTAMP WITH TIME ZONE NOT NULL,
> ```

Are these informational only since there is a "hard" ID reference to journey pattern in another database?



_Code quote:_
```TSQL
  observation_timestamp TIMESTAMP WITH TIME ZONE NOT NULL,
  snapshot_timestamp TIMESTAMP WITH TIME ZONE NOT NULL
```
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 22 at r11](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NFE_ugtEZwcF_3E_yjV:-NFE_ugtEZwcF_3E_yjW:bfeneri) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a8e16aa0a21cc91747c19d97bff5d2708c29c825/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L22)):*
> ```TSQL
> COMMENT ON SCHEMA service_pattern IS 'The service pattern model adapted from Transmodel: https://www.transmodel-cen.eu/model/index.htm?goto=2:3:4:723 ';
> 
> CREATE TABLE service_pattern.scheduled_stop_point_in_journey_pattern_ref (
> ```

There is no sequence number (order) column in this table. Stop points are in undefined order here. I assume we need order field, don't we? E.g. to verify and guarantee that arrival/departure times in `passing_times.timetabled_passing_time` are in correct order when compared to the stop points appearing in the original referenced journey pattern.



_Code quote:_
```TSQL
service_pattern.scheduled_stop_point_in_journey_pattern_ref
```

jschummer (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: all files reviewed, 8 unresolved discussions (waiting on @jarkkoka and @pallost)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 9 at r11](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NFEZvC-DKzVH2dt0BT6:-NFHxNwt-Yg3MPs4UcE6:bjjh1bo) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a8e16aa0a21cc91747c19d97bff5d2708c29c825/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L9)):*

Are these informational only since there is a "hard" ID reference to journey pattern in another database?
</blockquote></details>

These can in the future let us check if the journey pattern has changed on the routes & lines -side after an export was done (i.e. this `journey_pattern.journey_pattern_ref` was created).

The "hard" ID reference does not affect this, since the ID will stay the same, even if the journey pattern's contents change.
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 22 at r11](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NFE_ugtEZwcF_3E_yjV:-NFHxtbMBDHaYDN5BhGM:b-rd0rgj) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a8e16aa0a21cc91747c19d97bff5d2708c29c825/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L22)):*

There is no sequence number (order) column in this table. Stop points are in undefined order here. I assume we need order field, don't we? E.g. to verify and guarantee that arrival/departure times in `passing_times.timetabled_passing_time` are in correct order when compared to the stop points appearing in the original referenced journey pattern.
</blockquote></details>

Now that I start to think about it, I think you have a good point. Maybe we need to duplicate this information from the routes & lines -side?

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: all files reviewed, 8 unresolved discussions (waiting on @jschummer and @pallost)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 9 at r11](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NFEZvC-DKzVH2dt0BT6:-NFI2JS_0BTv1EiClEsI:bkq973u) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a8e16aa0a21cc91747c19d97bff5d2708c29c825/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L9)):*

These can in the future let us check if the journey pattern has changed on the routes & lines -side after an export was done (i.e. this `journey_pattern.journey_pattern_ref` was created).

The "hard" ID reference does not affect this, since the ID will stay the same, even if the journey pattern's contents change.
</blockquote></details>

Ok, got it.

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: all files reviewed, 9 unresolved discussions (waiting on @jarkkoka and @jschummer)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 95 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NDRXyYX0aDjJUfPea6O:-NFEBQEw5-f3pR4OEAXd:b-747qab) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/ed43ac77559b9abedd36e5b5389fda7841f5a5a6/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L95)):*

I think they definitely need to be modelled when we import data from Hastus. But I agree that this is not necessarily something that needs to be done in this PR.
</blockquote></details>

let's not then bring it into scope for this PR :+1: I created a separate ticket about this: https://github.com/HSLdevcom/jore4/issues/1006
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 121 at r11](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NFIZ-Dp2sg5KJpv_83X:-NFIZ-Dp2sg5KJpv_83Y:blpfdgh) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a8e16aa0a21cc91747c19d97bff5d2708c29c825/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L121)):*
> ```TSQL
>   scheduled_stop_point_in_journey_pattern_ref_id UUID REFERENCES service_pattern.scheduled_stop_point_in_journey_pattern_ref (scheduled_stop_point_in_journey_pattern_ref_id) NOT NULL,
>   arrival_time INTERVAL NULL,
>   departure_time INTERVAL NOT NULL
> ```

as agreed, let's make both arrival_time and departure_time nullable and create a generated column that is always not nullable. Should also create a constraint that either arrival or departure time must be defined

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: 38 of 40 files reviewed, 8 unresolved discussions (waiting on @jarkkoka and @jschummer)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 9 at r11](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NFEZvC-DKzVH2dt0BT6:-NFJJx3b3rgKg0yib_Le:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a8e16aa0a21cc91747c19d97bff5d2708c29c825/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L9)):*

Ok, got it.
</blockquote></details>

Done.
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 22 at r11](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NFE_ugtEZwcF_3E_yjV:-NFJK06e1jUSxePDY8iQ:b150dml) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a8e16aa0a21cc91747c19d97bff5d2708c29c825/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L22)):*

Now that I start to think about it, I think you have a good point. Maybe we need to duplicate this information from the routes & lines -side?
</blockquote></details>

Good point, I should add it. I created a separate ticket about implementing the constraint for having the passing times in the correct order: https://github.com/HSLdevcom/jore4/issues/1015
let's not take that constraint into the scope of this PR.
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 121 at r11](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NFIZ-Dp2sg5KJpv_83X:-NFNI2Cy5u88aNengtJm:b-896fix) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a8e16aa0a21cc91747c19d97bff5d2708c29c825/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L121)):*

as agreed, let's make both arrival_time and departure_time nullable and create a generated column that is always not nullable. Should also create a constraint that either arrival or departure time must be defined
</blockquote></details>

Done.

jarkkoka (approved):
> Reviewed 2 of 2 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: all files reviewed, 6 unresolved discussions (waiting on @jschummer and @pallost)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 22 at r11](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NFE_ugtEZwcF_3E_yjV:-NFTDKvyAh4fQImAtoNZ:b2nhb2c) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/a8e16aa0a21cc91747c19d97bff5d2708c29c825/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L22)):*

Good point, I should add it. I created a separate ticket about implementing the constraint for having the passing times in the correct order: https://github.com/HSLdevcom/jore4/issues/1015
let's not take that constraint into the scope of this PR.
</blockquote></details>

Ack
___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 126 at r12](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NFTDt2EAF8YS4Bazsqz:-NFTDt2EAF8YS4Bazsr-:b-2mox2k) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/785cd885fdba3307c3627c7e1702e709a4110262/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L126)):*
> ```TSQL
>     CASE
>       WHEN departure_time IS NULL THEN arrival_time
>       ELSE departure_time
> ```

More concise way of doing this:

```sql
coalesce(departure_time, arrival_time)
```



_Code quote:_
```TSQL
    CASE
      WHEN departure_time IS NULL THEN arrival_time
      ELSE departure_time
    END
```

jarkkoka (changes requested):
> Reviewed 2 of 2 files at r13, 1 of 1 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: all files reviewed, 7 unresolved discussions (waiting on @jschummer and @pallost)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 123 at r16](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96#-NFTLtB20OhuwrmM9b02:-NFTLtB20OhuwrmM9b03:b7ib2n0) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/618a7eea1a8b74519e9f50d6a927d27a169b14fb/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L123)):*
> ```TSQL
>   arrival_time interval NULL,
>   departure_time interval NULL,
>   passing_time interval NOT NULL GENERATED ALWAYS AS (COALESCE(departure_time, arrival_time)) STORED,
> ```

The convention is to put function calls as well in lowercase.



_Code quote:_
```TSQL
COALESCE
```

jarkkoka (approved):
> Reviewed 1 of 1 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/96)* status: all files reviewed, 7 unresolved discussions (waiting on @jschummer and @pallost)

PR #98, opened 2022-10-03
Validity time as date

jschummer:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98)


jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98)* status: 1 of 43 files reviewed, 4 unresolved discussions (waiting on @jschummer)

___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 6 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NE_tPDdDOyAAPlzh7IZ:-NE_vr0xC6Otm8ewVVlq:bfkgc7g) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L6)):*

The convention in the migration scripts (starting from the initial public transport schema) is to write data types with lowercase characters. In this PR, data types are in uppercase. I'd stick to the original convention.
</blockquote></details>

Actually, this PR mixes both lowercase and uppercase data types.

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98)* status: 1 of 43 files reviewed, 6 unresolved discussions (waiting on @jschummer)

___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 144 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NE_yB-Y8cHA9TE6cKLx:-NE_yB-Y8cHA9TE6cKLy:bfdk68e) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L144)):*
> ```TSQL
> COMMENT ON COLUMN
>   service_pattern.scheduled_stop_point.validity_start IS
>   'The point in time when the stop becomes valid. If NULL, the stop has been always valid.';
> ```

"The day when stop becomes valid." or something similar. "point in time" is not a good description since data type change: `timestamp` -> `date`.



_Code quote:_
```TSQL
The point in time when the stop becomes valid.
```
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 147 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NE_xpC4A8BXtckP8b5I:-NE_xpC5-ugsa2CZsSPV:b-sppn82) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L147)):*
> ```TSQL
> COMMENT ON COLUMN
>   service_pattern.scheduled_stop_point.validity_end IS
>   'The point in time from which onwards the stop is no longer valid. If NULL, the stop will be always valid.';
> ```

"The last day when stop is valid." or something similar. "point in time" is not a good description since data type change: `timestamp` -> `date`.



_Code quote:_
```TSQL
The point in time from which onwards the stop is no longer valid
```

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98)* status: 1 of 43 files reviewed, 9 unresolved discussions (waiting on @jschummer)

___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 51 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa1Zw267FOjgrVUKJn:-NEa1Zw267FOjgrVUKJo:b-c6n5ht) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L51)):*
> ```TSQL
>   Note that the range returned by this function will work well for calculating overlaps and merging,
>   but does not contain the correct logical upper bound. In order to calculate the upper bound of a range
>   created by this function, use the internal_utils.upperinc function.';
> ```

Reference to an `upperinc` not found in the function body.



_Code quote:_
```TSQL
internal_utils.upperinc
```
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 68 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa1kfE5dUOKyER5V9O:-NEa1kfE5dUOKyER5V9P:b219myn) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L68)):*
> ```TSQL
>   )
>   IS 'This function calculates the upper bound of a date range created by the
>   internal_utils.daterange_upperinc function.';
> ```

It seems to me that this function does not exist in this PR. Was it introduced in an earlier PR or is it renamed?



_Code quote:_
```TSQL
daterange_upperinc
```
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 522 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa2-VA9u0HLhHOWCfA:-NEa2-VA9u0HLhHOWCfB:b219myn) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L522)):*
> ```TSQL
> SELECT (SELECT array_agg(DISTINCT label) FROM route_param) AS labels,
>        lower(validity_range)                               AS validity_start,
>        -- since the upper bound was extended by the internal_utils.daterange_upperinc function above, we need
> ```

It seems to me that this function does not exist in this PR. Was it introduced in an earlier PR or is it renamed?



_Code quote:_
```TSQL
 internal_utils.daterange_upperinc
```

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98)* status: 1 of 43 files reviewed, 10 unresolved discussions (waiting on @jschummer)

___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 134 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa2QUW8rLpFK2a2lF6:-NEa2QUW8rLpFK2a2lF7:b-o7meb9) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L134)):*
> ```TSQL
> COMMENT ON COLUMN
>   service_pattern.scheduled_stop_point.direction IS
>   'The direction(s) of traffic with respect to the digitization, i.e. the direction of the specified line string.';
> ```

`LineString` is an actual geometrical type.



_Code quote:_
```TSQL
line string
```

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98)* status: 1 of 43 files reviewed, 11 unresolved discussions (waiting on @jschummer)

___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 645 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa5bXiEexMUlC31NYw:-NEa5bXiEexMUlC31NYx:b-c5ug5f) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L645)):*
> ```TSQL
>   -- form the list of potential validity span boundaries
>   -- e.g.
>   -- id, validity_start, is_start, key, prio
> ```

`id` seems not to be part of the query results.



_Code quote:_
```TSQL
-- id,
```

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98)* status: 1 of 43 files reviewed, 14 unresolved discussions (waiting on @jschummer)

___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 665 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa5xef2b-ZzBRL0t4x:-NEa5xeg8sZUio2jXlfz:b-14mm45) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L665)):*
> ```TSQL
>   ------------------------------------------------------------------
>   --  1,     2020-01-01,     true,   A,   10,           1,         4
>   --  2,     2022-01-01,     true,   A,   20,           2,         3
> ```

It seems to me that `id` column is not included in the actual results of the `ordered_boundary` query.



_Code quote:_
```TSQL
  -- id, validity_start, is_start, key, prio, start_order, end_order
  ------------------------------------------------------------------
  --  1,     2020-01-01,     true,   A,   10,           1,         4
  --  2,     2022-01-01,     true,   A,   20,           2,         3
  --  1,     2025-01-01,    false,   A,   10,           3,         2
  --  2,     2027-01-01,    false,   A,   20,           4,         1
```
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 685 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa6Lep6kJ00DzN9lLR:-NEa6Lep6kJ00DzN9lLS:b-f2vyte) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L685)):*
> ```TSQL
>   --  1,     2020-01-01,     true,   A,   10,           1,         4,                 10
>   --  2,     2022-01-01,     true,   A,   20,           2,         3,                 20
>   --  1,     2025-01-01,    false,   A,   10,           3,         2,                 20
> ```

It seems to me that `id` column is not included in the actual results of the `marked_min_start_priority` query.



_Code quote:_
```TSQL
  -- id, validity_start, is_start, key, prio, start_order, end_order, cur_start_priority
  --------------------------------------------------------------------------------------
  --  1,     2020-01-01,     true,   A,   10,           1,         4,                 10
  --  2,     2022-01-01,     true,   A,   20,           2,         3,                 20
  --  1,     2025-01-01,    false,   A,   10,           3,         2,                 20
  --  2,     2027-01-01,    false,   A,   20,           4,         1,                  0
```
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 715 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa6UVa61YvWe4sh8Eu:-NEa6UVa61YvWe4sh8Ev:b-xx7cu2) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L715)):*
> ```TSQL
>   --------------------------------------------------------------------------------------------------------
>   --  2,     2027-01-01,    false,   A,   20,           4,         1,                  0,               20
>   --  1,     2025-01-01,    false,   A,   10,           3,         2,                 20,               20
> ```

Same comment as previously with regard to the `id` column.



_Code quote:_
```TSQL
  -- id, validity_start, is_start, key, prio, start_order, end_order, cur_start_priority, cur_end_priority
  --------------------------------------------------------------------------------------------------------
  --  2,     2027-01-01,    false,   A,   20,           4,         1,                  0,               20
  --  1,     2025-01-01,    false,   A,   10,           3,         2,                 20,               20
  --  2,     2022-01-01,     true,   A,   20,           2,         3,                 20,                0
  --  1,     2020-01-01,     true,   A,   10,           1,         4,                 10,               10
```

jschummer (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98)* status: 0 of 43 files reviewed, 15 unresolved discussions (waiting on @jarkkoka)

___
*[`-- commits` line 7 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NE_cpXzDNjYbEWXJIU2):*

"GraphQL API" since "corresponds to the way UI..." already means user.
</blockquote></details>

Fixed.
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 5 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NE_ujle2c9CcZD9PdA8:-NFOHPgo1c0hGTRx869Q:bop42dv) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L5)):*

IMO "next_day" would be much more concise and understandable name for this function. Adding context information to the function name is not a good approach for such a simple function as this is.
</blockquote></details>

You have a good point, I renamed the functions to "next_day" and "prev_day".
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 6 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NE_tPDdDOyAAPlzh7IZ:-NFOIdVE91AR_xV2eGzl:b-if1715) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L6)):*

Actually, this PR mixes both lowercase and uppercase data types.
</blockquote></details>

That's because the function definitions were copied from other separate migrations, which used slightly different styles. I now changed all types to use lower case.

In the long term, we might want to think about some automatic formatting solution.
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 17 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NE_v6Q47mZoUt-2h7wy:-NFOIz_49E5I4a6fn1yA:bz97mll) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L17)):*

IMO "previous_day" would be much more concise and understandable name for this function. Adding context information to the function name is not a good approach for such a simple function as this is.
</blockquote></details>

I now changed this to "prev_day".
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 51 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa1Zw267FOjgrVUKJn:-NFOJ1MJ7kbd2xxQH9Fm:b7kplag) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L51)):*

Reference to an `upperinc` not found in the function body.
</blockquote></details>

Nice catch. This was a working name for one of the functions defined above. I didn't notice that idea's refactoring tool didn't obviously take these comment mentions into account. Fixed it.
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 68 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa1kfE5dUOKyER5V9O:-NFOJIAM8i9rN1jVylrx:b-uu8tfs) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L68)):*

It seems to me that this function does not exist in this PR. Was it introduced in an earlier PR or is it renamed?
</blockquote></details>

This was a working name for one of the functions defined above. I didn't notice that idea's refactoring tool didn't obviously take these comment mentions into account. Fixed it.
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 134 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa2QUW8rLpFK2a2lF6:-NFOJKEAB79ILglxoWGv:b-tag63h) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L134)):*

`LineString` is an actual geometrical type.
</blockquote></details>

Since this is copied from a previous migration, and is not supposed to be changed by this PR, I'm not sure this is the right place to fix it.

Could we get back to this kind of fixes when we found a better migration mechanism, which let's us make this kind of changes more easily?
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 144 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NE_yB-Y8cHA9TE6cKLx:-NFOJaVbEkhtQ0oAMGQh:b-hc3v2n) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L144)):*

"The day when stop becomes valid." or something similar. "point in time" is not a good description since data type change: `timestamp` -> `date`.
</blockquote></details>

I think a date can also depict a point in time, it just uses a different resolution than a (ms-resolution) timestamp.

But despite this, you are probably right that it may be clearer to use "date" here. Changed it.
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 147 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NE_xpC4A8BXtckP8b5I:-NFOK3lxAKFJTheFXCtQ:buyydde) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L147)):*

"The last day when stop is valid." or something similar. "point in time" is not a good description since data type change: `timestamp` -> `date`.
</blockquote></details>

Fixed it.
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 522 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa2-VA9u0HLhHOWCfA:-NFOK75X15zVXKZnMJ78:byr3z03) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L522)):*

It seems to me that this function does not exist in this PR. Was it introduced in an earlier PR or is it renamed?
</blockquote></details>

Same as above, relict from a previous version of this migration. Fixed it.
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 645 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa5bXiEexMUlC31NYw:-NFOKBnRBEuKKj5ymWj4:b-wyya7y) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L645)):*

`id` seems not to be part of the query results.
</blockquote></details>

I included the id just to make it easier for the reader to see, which table row the returned row originates from.

I'd rather not touch this here, since it is just copied from another migration and not supposed to be changed here.

Like said above, maybe we can get back to these when we have a better migration mechanism?
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 665 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa5xef2b-ZzBRL0t4x:-NFOKMd18fnh7lCpAhga:b-7q0uej) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L665)):*

It seems to me that `id` column is not included in the actual results of the `ordered_boundary` query.
</blockquote></details>

Same as above, I included the id just to make it easier for the reader to see, which table row the returned row originates from.

I'd rather not touch this here, since it is just copied from another migration and not supposed to be changed here.

Like said above, maybe we can get back to these when we have a better migration mechanism?
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 685 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa6Lep6kJ00DzN9lLR:-NFOKPsNBKOJtOgY2tvj:b-7q0uej) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L685)):*

It seems to me that `id` column is not included in the actual results of the `marked_min_start_priority` query.
</blockquote></details>

Same as above, I included the id just to make it easier for the reader to see, which table row the returned row originates from.

I'd rather not touch this here, since it is just copied from another migration and not supposed to be changed here.

Like said above, maybe we can get back to these when we have a better migration mechanism?
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 715 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa6UVa61YvWe4sh8Eu:-NFOKk5157PSgLS8LoVG:b-7q0uej) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L715)):*

Same comment as previously with regard to the `id` column.
</blockquote></details>

Same as above, I included the id just to make it easier for the reader to see, which table row the returned row originates from.

I'd rather not touch this here, since it is just copied from another migration and not supposed to be changed here.

Like said above, maybe we can get back to these when we have a better migration mechanism?
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 759 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa73p237yMa6LVxWwx:-NFOKq8J4tOmgs-qi0Fx:b-n4fmgb) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L759)):*

Extra indend before `WINDOW` since it is not part of `WHERE` predicates.
</blockquote></details>

Idea's formatting tool seems to think this is the way to do it, but you're of course right, not logical at all. Fixed it.

jarkkoka (commented):
> Reviewed 41 of 43 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98)* status: all files reviewed, 6 unresolved discussions (waiting on @jschummer)

___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 134 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa2QUW8rLpFK2a2lF6:-NFObJDZ10d1s6sCISp8:bqyrjka) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L134)):*

Since this is copied from a previous migration, and is not supposed to be changed by this PR, I'm not sure this is the right place to fix it.

Could we get back to this kind of fixes when we found a better migration mechanism, which let's us make this kind of changes more easily?
</blockquote></details>

Ok. Although this is quite easy to forget to do. :)
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 665 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa5xef2b-ZzBRL0t4x:-NFORdA4ETYcy4AjvDkt:bv2rclh) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L665)):*

Same as above, I included the id just to make it easier for the reader to see, which table row the returned row originates from.

I'd rather not touch this here, since it is just copied from another migration and not supposed to be changed here.

Like said above, maybe we can get back to these when we have a better migration mechanism?
</blockquote></details>

Ok. Although I'd note that those differences (to the preceding `up.sql`) would be of little concern in a proper diff tool. There are already a lot of small changes due to uppercase to lowercase conversions done for data types and function names but even those changes are trivial and fairly unobtrusive in a proper diff tool such as macOS's FileMerge that I used to compare this `up.sql` to the one in the folder of the preceding migration.

What really makes diffing hard is that this PR is not an update to just one migration script but several of them. So, there are quite a lot of changes made here anyway that cannot be easily detected line by line (in the version history or by using a diff tool) or tracked back to previous states of the database.
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 49 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NFOfYrt8YBR2x-8Py6t:-NFOfYrt8YBR2x-8Py6u:b-wg8i0d) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/1f6e0aec0f282e42673efd98324e001458101197/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L49)):*
> ```TSQL
>   a daterange with the upper bound interpreted as a closed bound. This function behaves the same way
>   as "daterange(lower_bound, upper_bound, ''[]'')", but because we also need to access the bounds (and
>   this would anyway require manual addition / subtraction of 1 day), we rather do all processing in
> ```

I wouldn't say this is wrong approach, but I think for many people (including myself) sticking to standard functions makes more sense and code more understandable.



_Code quote:_
```TSQL
  this would anyway require manual addition / subtraction of 1 day), we rather do all processing in
  custom function
```
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 355 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NFOcI3ZE6XIEwhht4g-:-NFOcI3_9POmZW3EIRgg:b-u2i90x) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/1f6e0aec0f282e42673efd98324e001458101197/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L355)):*
> ```TSQL
>           supported_vehicle_mode);
> END;
> $$ LANGUAGE plpgsql;
> ```

Looks like plain SQL but can be a misinterpretation from my part.



_Code quote:_
```TSQL
 plpgsql
```
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 614 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NFOU0Ua60TdaeYk1f6x:-NFOU0Ua60TdaeYk1f6y:b1vt28h) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/1f6e0aec0f282e42673efd98324e001458101197/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L614)):*
> ```TSQL
>           internal_utils.daterange_closed_upper(filter_validity_start, filter_validity_end)
>       AND (r.label = ANY (filter_route_labels) OR filter_route_labels IS NULL)
>       AND (r.priority < upper_priority_limit OR upper_priority_limit IS NULL)
> ```

This parameter could be renamed to `upper_priority_limit_exclusive` or something similar expressing the exclusivity.

It would in line with all the extra work (internal functions) done for treating upper bounds of `daterange`s.

Maybe to be done in the subsequent PR's if the differences are desired to be kept to minimum. 



_Code quote:_
```TSQL
upper_priority_limit
```
___
*[`test/hasura/util/dataset.ts` line 62 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NFOfMsN8po0-gxb5WfS:-NFOfMsN8po0-gxb5WfT:b7laxn4) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/1f6e0aec0f282e42673efd98324e001458101197/test/hasura/util/dataset.ts#L62)):*
> ```XML
>     );
> 
> export const asGraphQlTimestampObject = (obj: {
> ```

Maybe should be renamed to `asGraphQlDateObject`.



_Code quote:_
```XML
asGraphQlTimestampObject
```

jschummer (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98)* status: 42 of 43 files reviewed, 6 unresolved discussions (waiting on @jarkkoka)

___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 134 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa2QUW8rLpFK2a2lF6:-NFRwDwB4Ee7353pgutt:b89ddzf) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L134)):*

Ok. Although this is quite easy to forget to do. :)
</blockquote></details>

Yes, of course :grimacing: 

But then, we also have the card about flattening the migrations (https://github.com/HSLdevcom/jore4/issues/923), which would be natural to do at the same time with changing the migration tool. So with that rewriting a general proof-reading could be performed.

Anyhow, I suppose there is not much harm in correcting this now, so I changed it.
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 665 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa5xef2b-ZzBRL0t4x:-NFRxq4qCdnVt1uca7gA:b-rv2a6p) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L665)):*

Ok. Although I'd note that those differences (to the preceding `up.sql`) would be of little concern in a proper diff tool. There are already a lot of small changes due to uppercase to lowercase conversions done for data types and function names but even those changes are trivial and fairly unobtrusive in a proper diff tool such as macOS's FileMerge that I used to compare this `up.sql` to the one in the folder of the preceding migration.

What really makes diffing hard is that this PR is not an update to just one migration script but several of them. So, there are quite a lot of changes made here anyway that cannot be easily detected line by line (in the version history or by using a diff tool) or tracked back to previous states of the database.
</blockquote></details>

Yes, that is of course true.

Part of the problem with manual diffing in this case is that I have started to use IDEA automatic formatting (which doesn't always work really well, as you have noticed also in this PR), only some migrations ago. So the older migration code, which is referenced here, has a formatting that may be quite different.

 I suppose that latest with this PR, we have justification to use a different migration mechanism in the future...
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 49 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NFOfYrt8YBR2x-8Py6t:-NFRu7554X3sEMC87bXT:b-cxrh2y) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/1f6e0aec0f282e42673efd98324e001458101197/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L49)):*

I wouldn't say this is wrong approach, but I think for many people (including myself) sticking to standard functions makes more sense and code more understandable.
</blockquote></details>

In general, I totally agree. But here we have the choice between using the following sets of code:

A)
`daterange(lower_bound, upper_bound, '[]')` for constructing the daterange and
`prev_day(upper(my_date_range))` for extracting the upper bound

and B)
`daterange(lower_bound, next_day(upper_bound))` for constructing the daterange and
`prev_day(upper(my_date_range))` for extracting the upper bound 

Even though A uses a standard function for constructing the daterange, we have to use custom processing (subtracting 1 day) to get the real upper bound out of the daterange. On first glimpse, A) might even look like a bug to me. That's why I thought it's better to explicitly add a day to the upper bound and construct the daterange as ''(]'', just the way postgresql handles it internally.
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 355 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NFOcI3ZE6XIEwhht4g-:-NFRx7I5Bp20dp3oj5hh:b-4ug4c2) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/1f6e0aec0f282e42673efd98324e001458101197/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L355)):*

Looks like plain SQL but can be a misinterpretation from my part.
</blockquote></details>

It's two SQL statements, so I suppose the whole block doesn't qualify as plain SQL.
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 614 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NFOU0Ua60TdaeYk1f6x:-NFRxXz87Orv_dzJeRfq:bxzo04h) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/1f6e0aec0f282e42673efd98324e001458101197/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L614)):*

This parameter could be renamed to `upper_priority_limit_exclusive` or something similar expressing the exclusivity.

It would in line with all the extra work (internal functions) done for treating upper bounds of `daterange`s.

Maybe to be done in the subsequent PR's if the differences are desired to be kept to minimum. 
</blockquote></details>

Yes, I'd prefer it if this could be done in another context. And of course you are right, it would be clearer with the `exclusive` added to it.
___
*[`test/hasura/util/dataset.ts` line 62 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NFOfMsN8po0-gxb5WfS:-NFRyNkx7g1B4y5EdF1-:b-5962px) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/1f6e0aec0f282e42673efd98324e001458101197/test/hasura/util/dataset.ts#L62)):*

Maybe should be renamed to `asGraphQlDateObject`.
</blockquote></details>

Good catch! Fixed it.

jarkkoka (approved):
:lgtm: 

> Reviewed 31 of 31 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-:-NFSLCeeC-Hf3C3u7jVc:bg7igkv)* status: all files reviewed, 4 unresolved discussions (waiting on @jschummer)

___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 134 at r1](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NEa2QUW8rLpFK2a2lF6:-NFSI4lI-ZJXzk_8GA3m:b-54avj1) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/0bec28d8541764a949bd96e9c69166d0a8b5c5d5/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L134)):*

Yes, of course :grimacing: 

But then, we also have the card about flattening the migrations (https://github.com/HSLdevcom/jore4/issues/923), which would be natural to do at the same time with changing the migration tool. So with that rewriting a general proof-reading could be performed.

Anyhow, I suppose there is not much harm in correcting this now, so I changed it.
</blockquote></details>

Ok, thanks!
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 49 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NFOfYrt8YBR2x-8Py6t:-NFSKUInApSycrh_esGq:b-l5sy98) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/1f6e0aec0f282e42673efd98324e001458101197/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L49)):*

In general, I totally agree. But here we have the choice between using the following sets of code:

A)
`daterange(lower_bound, upper_bound, '[]')` for constructing the daterange and
`prev_day(upper(my_date_range))` for extracting the upper bound

and B)
`daterange(lower_bound, next_day(upper_bound))` for constructing the daterange and
`prev_day(upper(my_date_range))` for extracting the upper bound 

Even though A uses a standard function for constructing the daterange, we have to use custom processing (subtracting 1 day) to get the real upper bound out of the daterange. On first glimpse, A) might even look like a bug to me. That's why I thought it's better to explicitly add a day to the upper bound and construct the daterange as ''(]'', just the way postgresql handles it internally.
</blockquote></details>

Ok. It seem to me that you prioritise consistency of code style here. Like I said earlier, it's not a wrong approach. So, I'm fine with it. But we should stick to this style and convention in the future as well.
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 355 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NFOcI3ZE6XIEwhht4g-:-NFSISGK1-XAqL7hCFy1:byc62vd) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/1f6e0aec0f282e42673efd98324e001458101197/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L355)):*

It's two SQL statements, so I suppose the whole block doesn't qualify as plain SQL.
</blockquote></details>

Ok. It seems so.
___
*[`migrations/generic/default/1664191395447_validity_time_as_date/up.sql` line 614 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/98#-NFOU0Ua60TdaeYk1f6x:-NFSKFnt45os3ipCwSmH:b2nhll9) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/1f6e0aec0f282e42673efd98324e001458101197/migrations/generic/default/1664191395447_validity_time_as_date/up.sql#L614)):*

Yes, I'd prefer it if this could be done in another context. And of course you are right, it would be clearer with the `exclusive` added to it.
</blockquote></details>

Ok.

PR #99, opened 2022-10-11
Add priority to timetables

pallost:
For now, we only define priority on the timetable (VEHICLE SCHEDULE FRAME) level. This affects all contained entities within (VEHICLE SERVICE, VEHICLE JOURNEY, etc).
Timetable priorities are evaluated on an OPERATING DAY basis: if on the given OPERATING DAY another active, higher priority version is found, then that timetable will take precedence for that given day.

- - -
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/99)

Closes https://github.com/HSLdevcom/jore4/issues/1003

### DIFF
--- a/migrations/generic/timetables/1663231564515_timetables_initial/up.sql
+++ b/migrations/generic/timetables/1663231564515_timetables_initial/up.sql
@@ -67,12 +67,14 @@ CREATE TABLE vehicle_schedule.vehicle_schedule_frame (
   vehicle_schedule_frame_id uuid DEFAULT gen_random_uuid() PRIMARY KEY,
   name_i18n jsonb NOT NULL,
   validity_start date NULL,
-  validity_end date NULL
+  validity_end date NULL,
+  priority int NOT NULL
 );
 COMMENT ON TABLE vehicle_schedule.vehicle_schedule_frame IS 'A coherent set of BLOCKS, COMPOUND BLOCKs, COURSEs of JOURNEY and VEHICLE SCHEDULEs to which the same set of VALIDITY CONDITIONs have been assigned. Transmodel: https://www.transmodel-cen.eu/model/index.htm?goto=3:7:2:993 ';
 COMMENT ON COLUMN vehicle_schedule.vehicle_schedule_frame.name_i18n IS 'Human-readable name for the VEHICLE SCHEDULE FRAME';
 COMMENT ON COLUMN vehicle_schedule.vehicle_schedule_frame.validity_start IS 'OPERATING DAY when the VEHICLE SCHEDULE FRAME validity starts. Null if always has been valid.';
 COMMENT ON COLUMN vehicle_schedule.vehicle_schedule_frame.validity_end IS 'OPERATING DAY when the VEHICLE SCHEDULE FRAME validity end. Null if always will be valid.';
+COMMENT ON COLUMN vehicle_schedule.vehicle_schedule_frame.priority IS 'The priority of the timetable definition. The definition may be overridden by higher priority definitions.';
 
 -------------------- Vehicle Service --------------------
 
@@ -84,7 +86,7 @@ CREATE TABLE vehicle_service.vehicle_service (
   day_type_id uuid REFERENCES service_calendar.day_type (day_type_id) NOT NULL,
   vehicle_schedule_frame_id uuid REFERENCES vehicle_schedule.vehicle_schedule_frame (vehicle_schedule_frame_id) NOT NULL
 );
-COMMENT ON TABLE vehicle_service.vehicle_service IS 'A work plan for a single vehicle for a whole day, planned for a specific DAY TYPE. A VEHICLE SERVICE includes one or several VEHICLE SERVICE PARTs. Transmodel: https://www.transmodel-cen.eu/model/index.htm?goto=3:5:965 ';
+COMMENT ON TABLE vehicle_service.vehicle_service IS 'A work plan for a single vehicle for a whole day, planned for a specific DAY TYPE. A VEHICLE SERVICE includes one or several BLOCKs. If there is no service on a given day, it does not include any BLOCKs. Transmodel: https://www.transmodel-cen.eu/model/index.htm?goto=3:5:965 ';
 COMMENT ON COLUMN vehicle_service.vehicle_service.day_type_id IS 'The DAY TYPE for the VEHICLE SERVICE.';
 COMMENT ON COLUMN vehicle_service.vehicle_service.vehicle_schedule_frame_id IS 'Human-readable name for the VEHICLE SCHEDULE FRAME';
 
--- a/migrations/seed-data/timetables/2524600900000_seed-timetables/up.sql
+++ b/migrations/seed-data/timetables/2524600900000_seed-timetables/up.sql
@@ -20,9 +20,9 @@ VALUES
 ON CONFLICT DO NOTHING;
 
 INSERT INTO vehicle_schedule.vehicle_schedule_frame
-  (vehicle_schedule_frame_id, name_i18n, validity_start, validity_end)
+  (vehicle_schedule_frame_id, name_i18n, validity_start, validity_end, priority)
 VALUES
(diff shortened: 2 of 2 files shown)


jarkkoka (commented):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/99)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @jschummer)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 71 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/99#-NEK9AxcDrrg-ksF_z-g:-NFNURSi7YKYe3KMMdxP:b-4zlmxr) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/495743999101650ba25e3723727d7fdef2809732/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L71)):*

A check constraint should also be added to enforce unique priorities with non-overlapping validity periods. I suppose this will follow (possibly with other constraints?) in a separate PR?

When adding such a constraint, note that this cannot be handled using the postgresql `daterange` function, because it always creates a range that has the upper bound open and we want the upper bound to be closed.

I added a helper function for this in the timestamp<->daterange PR (https://reviewable.io/reviews/HSLdevcom/jore4-hasura/98#-): `internal_utils.daterange_closed_upper`. But of course that cannot be used as such, because it's in a different DB...
</blockquote></details>

I don't see a reason why we could not use `daterange` function for constraints. Actually, IMO `daterange` function would suite neatly for implementing a **unique** constraint for pairs of validity ranges and priorities. Like in an example below. In the example `daterange` is used only inside the constraint. Of course, the daterange could be added also as a computed field into the table definition itself.

```sql
CREATE EXTENSION btree_gist;

CREATE TABLE test (
  id uuid DEFAULT gen_random_uuid() PRIMARY KEY,
  validity_start date,
  validity_end date,
  priority int NOT NULL
);

ALTER TABLE test 
  ADD CONSTRAINT unique_daterange_per_priority
    EXCLUDE USING gist (
      priority WITH =,
      daterange(validity_start, validity_end, '[]') WITH &&
    );

INSERT INTO test (validity_start, validity_end, priority) VALUES ('2022-10-27', '2022-10-27', 1);
INSERT INTO test (validity_start, validity_end, priority) VALUES ('2022-10-28', '2022-10-31', 1);
INSERT INTO test (validity_start, validity_end, priority) VALUES ('2022-10-27', '2022-10-28', 2);
```



_Code quote:_
```TSQL
 UUID DEFAULT gen_random_uuid() PRIMARY KEY
```

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/99)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @jarkkoka and @jschummer)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 71 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/99#-NEK9AxcDrrg-ksF_z-g:-NFNY6i09Ng3iKujG0EM:b4qfec8) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/495743999101650ba25e3723727d7fdef2809732/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L71)):*

I don't see a reason why we could not use `daterange` function for constraints. Actually, IMO `daterange` function would suite neatly for implementing a **unique** constraint for pairs of validity ranges and priorities. Like in an example below. In the example `daterange` is used only inside the constraint. Of course, the daterange could be added also as a computed field into the table definition itself.

```sql
CREATE EXTENSION btree_gist;

CREATE TABLE test (
  id uuid DEFAULT gen_random_uuid() PRIMARY KEY,
  validity_start date,
  validity_end date,
  priority int NOT NULL
);

ALTER TABLE test 
  ADD CONSTRAINT unique_daterange_per_priority
    EXCLUDE USING gist (
      priority WITH =,
      daterange(validity_start, validity_end, '[]') WITH &&
    );

INSERT INTO test (validity_start, validity_end, priority) VALUES ('2022-10-27', '2022-10-27', 1);
INSERT INTO test (validity_start, validity_end, priority) VALUES ('2022-10-28', '2022-10-31', 1);
INSERT INTO test (validity_start, validity_end, priority) VALUES ('2022-10-27', '2022-10-28', 2);
```
</blockquote></details>

This constraint actually needs a major work as it also depends on the journey pattern which is very deep in the data model. Thus, I wouldn't necessarily include it to this pull request. I detached it to its own ticket: https://github.com/HSLdevcom/jore4/issues/1016

jarkkoka (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/99)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @jschummer)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 71 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/99#-NEK9AxcDrrg-ksF_z-g:-NFNkK7LAuX5-O4UG4wU:bbn4j7o) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/495743999101650ba25e3723727d7fdef2809732/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L71)):*

This constraint actually needs a major work as it also depends on the journey pattern which is very deep in the data model. Thus, I wouldn't necessarily include it to this pull request. I detached it to its own ticket: https://github.com/HSLdevcom/jore4/issues/1016 
</blockquote></details>

It indeed seems that there is not enough information contained in this table based on which a unique constraint could be created.

jarkkoka (commented):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/99)* status: all files reviewed, 1 unresolved discussion (waiting on @jschummer)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 71 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/99#-NEK9AxcDrrg-ksF_z-g:-NFTFVZR--phegx9wfGN:b-3si04m) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/495743999101650ba25e3723727d7fdef2809732/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L71)):*

It indeed seems that there is not enough information contained in this table based on which a unique constraint could be created.
</blockquote></details>

I would remove the trailing comma after `priority` column definition. This should probably be corrected in a separate PR. What do you think?

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/99)* status: all files reviewed, 1 unresolved discussion (waiting on @jschummer)

___
*[`migrations/generic/timetables/1663231564515_timetables_initial/up.sql` line 71 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/99#-NEK9AxcDrrg-ksF_z-g:-NFTGGn41JuODUIDGmgp:bljigo) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/495743999101650ba25e3723727d7fdef2809732/migrations/generic/timetables/1663231564515_timetables_initial/up.sql#L71)):*

I would remove the trailing comma after `priority` column definition. This should probably be corrected in a separate PR. What do you think?
</blockquote></details>

Oops. There was missing something from the previous comment.

I meant to say that:
All the data types are in uppercase whereas the style convention of the project is to use lowercase characters for data types in table definitions. This should probably be corrected in a separate PR. What do you think?

jarkkoka (approved):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/99)* status: all files reviewed, 1 unresolved discussion (waiting on @jschummer)

PR #105, opened 2022-11-01
Add computed start_time/end_time fields to vehicle_journey table

tk-vincit:
This change is [Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/105)

Resolves HSLdevcom/jore4#1000


pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/hsldevcom/jore4-hasura/105)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @tk-vincit)

___
*[`migrations/generic/timetables/1667215510823_add_vehicle_journey_computed_fields/up.sql` line 3 at r2](https://reviewable.io/reviews/hsldevcom/jore4-hasura/105#-NFs-uFi52NWGRfKlM-b:-NFs-uFj5xsTo8fRB3PL:b-3rag92) ([raw file](https://github.com/hsldevcom/jore4-hasura/blob/9774c1292c6297919341bd62a7439dba07c4a67a/migrations/generic/timetables/1667215510823_add_vehicle_journey_computed_fields/up.sql#L3)):*
> ```TSQL
> -- NOTE: it would make more sense to return `interval` instead of `text`, but
> -- hasura won't let save those as computed fields:
> -- "Inconsistent object: in table "vehicle_journey.vehicle_journey": in computed field "start_time": the computed field "start_time" cannot be added to table "vehicle_journey.vehicle_journey" because the function "vehicle_journey.vehicle_journey_start_time" returning type interval is not a BASE type"
> ```

thanks for the comment!

PR #132, opened 2023-01-12
Move route variant to hsl schema

Leitsi:
Resolves https://github.com/HSLdevcom/jore4/issues/1098

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132)


pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132)* status: 0 of 104 files reviewed, 12 unresolved discussions (waiting on @eleksi and @Leitsi)

___
*[`README.md` line 245 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NL_WMc3E0GYt4l8XYay:-NL_WMc3E0GYt4l8XYaz:b-yw3t7k) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/README.md#L245)):*
> ```Markdown
>    `yarn test`
> 
> ### HSL schema specific tests
> ```

Thanks for documenting this
___
*[`migrations/hsl/default/1672839425530_add_hsl_route_variant/up.sql` line 6 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NL_YC7D3fh5zX-rm7Zh:-NL_YC7E89tfM2c3fzg7:b-la9mda) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/migrations/hsl/default/1672839425530_add_hsl_route_variant/up.sql#L6)):*
> ```TSQL
> COMMENT ON COLUMN route.route.variant IS 'The variant for route definition.';
> 
> ALTER TABLE route.route
> ```

we've had the CHECK constraints separately, as repeatable migrations. (In practice, just move this to a migration file in `migrations/hsl/default/2000000000101_R_after_migrate_route/up.sql`)
___
*[`migrations/hsl/default/1672910047335_include_variant_in_unique_label/up.sql` line 6 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NL_Ys0w1Zsy4hu1V_FJ:-NL_Ys0w1Zsy4hu1V_FK:bdsycco) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/migrations/hsl/default/1672910047335_include_variant_in_unique_label/up.sql#L6)):*
> ```TSQL
> 
> ALTER TABLE route.route
>   ADD COLUMN unique_label text GENERATED ALWAYS AS (label || '_' || coalesce(variant::TEXT,'')) STORED;
> ```

This might not look very nice when displayed when the variant attribute is NULL.
Should it be rather like this? 
`(label || (CASE WHEN variant IS NULL THEN '' ELSE '_' || variant::text END))`
___
*[`test/hasura/jest.config.js` line 20 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NL___I0-y4T4SC5T2G2:-NL___I0-y4T4SC5T2G3:bb7phvt) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/jest.config.js#L20)):*
> ```JavaScript
>   testPathIgnorePatterns: ['/node_modules/', hslTestRootFolder],
> };
> const hslCOnfig = {
> ```

typo
___
*[`test/hasura/hsl/datasets/factories.ts` line 7 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NLa5xUxDdFFopzNKp3B:-NLa5xUxDdFFopzNKp3C:b-bjtfp7) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/hsl/datasets/factories.ts#L7)):*
> ```XML
> import { HslTransportTarget, VehicleMode } from '@datasets-hsl/types';
> 
> // The generic data is typically from either factory methods or defaultDataset, which are different types, so using generics here in HSL converters.
> ```

these mappers from generic to HSL are not flexible at all. In fact I the current buildLine solution is not good either. Here's a suggestion for a quick fix that should allow flexibility but won't require you to refactor the whole codebase:

```
// Make given keys of a type required, leave the rest as-is
export type RequiredKeys<T, K extends keyof T> = Required<Pick<T, K>> &
  Omit<T, K>;

export const buildHSLLine = (input: Pick<Line, 'label' | 'vehicleMode'>) => {
  const {label, vehicleMode} = input;
  return {
  name_i18n: buildLocalizedString(`transport ${vehicleMode} line ${label}`),
  short_name_i18n: buildLocalizedString(`line ${label}`),
  primary_vehicle_mode: vehicleMode,
  type_of_line: buildTypeOfLine(vehicleMode),
  transport_target: HslTransportTarget.HelsinkiInternalTraffic,
  ...input,
  } as HSLLine;
})

```

(you may need to set some attributes of the Line type as optional)

This new buildHSLLine:
1. can be used to build lines based on label and vehicle mode
2. allows the caller to set other attributes too (if e.g. want to manually override the priority for some reason)
3. can be used as a mapping function (instead of the genericLineToHsl, you could just call this buildHSLLine with the same parameters)
___
*[`test/hasura/hsl/datasets/factories.ts` line 27 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NLa7lQZ6s7cH_Mpe57x:-NLa7lQ_7LDyYDjsvKTr:b-aym4tc) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/hsl/datasets/factories.ts#L27)):*
> ```XML
> };
> 
> export const genericRouteToHsl = <T>(genericRoute: T) => {
> ```

same as above, this function is not flexible enough
___
*[`test/hasura/hsl/datasets/setup.ts` line 2 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NLa7rk38PpbRUD1dmyZ:-NLa7rk38PpbRUD1dmy_:b-q7ho6u) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/hsl/datasets/setup.ts#L2)):*
> ```XML
> import * as genericSetup from '@datasets-generic/setup';
> import { defaultTableConfig } from '@datasets-hsl/defaultSetup';
> ```

should you rename this to "hslDefaultTableConfig" for clarity?
___
*[`test/hasura/hsl/datasets/types.ts` line 12 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NLa4_Cp1Z3Ltg-TTJM_:-NLa4_Cp1Z3Ltg-TTJMa:b-2igtmy) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/hsl/datasets/types.ts#L12)):*
> ```XML
> export enum HslTransportTarget {
>   HelsinkiInternalTraffic = 'helsinki_internal_traffic',
>   EspooAnd_kauniainen_internalTraffic = 'espoo_and_kauniainen_internal_traffic',
> ```

that's a weird looking enum key here :)
___
*[`test/hasura/hsl/datasets/types.ts` line 32 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NLa8O0hCIXAU3C18PUT:-NLa8O0hCIXAU3C18PUU:bd12t81) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/hsl/datasets/types.ts#L32)):*
> ```XML
> };
> 
> export const HslLineProps: Property[] = [
> ```

this is not a type but a constant. Let's rather have it as camelCase. Same for the other Props arrays
___
*[`test/hasura/hsl/datasets/defaultSetup/index.ts` line 11 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NLa8a2JDkyuUP0ahU01:-NLa8a2JDkyuUP0ahU02:b5g2g8z) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/hsl/datasets/defaultSetup/index.ts#L11)):*
> ```XML
> ];
> 
> const hslDefaultTableConfig: TableLikeConfig[] = genericDefaultTableConfig.map(
> ```

I have to say this is a quite ugly way to override the original values :D But let's leave it like this as the whole TableConfig shit is gonna be refactored anyway to separate the data from the schema properly
___
*[`test/hasura/generic/datasets/setup.ts` line 47 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NL_aCxJ3CIPk4WtJf8U:-NL_aCxJ3CIPk4WtJf8V:bzacaxa) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/a2667a89c3090133b0e65d419276c37fc81b667f/test/hasura/generic/datasets/setup.ts#L47)):*
> ```XML
>       );
> 
>       queryRunner = queryRunner.insertFromJson(
> ```

if you removed this, then you could also remove it from the other `queryRunner = queryRunner.query(...)` calls
___
*[`test/hasura/generic/util/db.ts` line 21 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NL_bgC96inpM-iSPUlg:-NL_bgC96inpM-iSPUlh:b-xo0m45) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/generic/util/db.ts#L21)):*
> ```XML
> 
>   insertFromJson = (table: string, jsonObjects: Record<string, unknown>[]) => {
>     jsonObjects.forEach((jsonObject) => {
> ```

this is not a very good coding pattern, it's not flexible at all. I'd rather like to see it implemented like this:
- refactoring the parameterSymbols calculation to its own function
- refactoring the value mapping to its own function

also, if it's a possibility, just take the props array directly from the TableConfig object instead of calculating the parameterSymbols. (it's available in the setupDb function where this function is called from)

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132)* status: 0 of 108 files reviewed, 11 unresolved discussions (waiting on @eleksi and @pallost)

___
*[`migrations/hsl/default/1672839425530_add_hsl_route_variant/up.sql` line 6 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NL_YC7D3fh5zX-rm7Zh:-NLe-EvNDf3GTm_kPD-k:b4c9lyr) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/migrations/hsl/default/1672839425530_add_hsl_route_variant/up.sql#L6)):*

we've had the CHECK constraints separately, as repeatable migrations. (In practice, just move this to a migration file in `migrations/hsl/default/2000000000101_R_after_migrate_route/up.sql`)
</blockquote></details>

:thumbsup: 
___
*[`migrations/hsl/default/1672910047335_include_variant_in_unique_label/up.sql` line 6 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NL_Ys0w1Zsy4hu1V_FJ:-NLdxClmCo07MwYcDT6L:b-bhzkqx) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/migrations/hsl/default/1672910047335_include_variant_in_unique_label/up.sql#L6)):*

This might not look very nice when displayed when the variant attribute is NULL.
Should it be rather like this? 
`(label || (CASE WHEN variant IS NULL THEN '' ELSE '_' || variant::text END))`
</blockquote></details>

True, not very pretty in that case, will change. This wasn't really intended to be shown in any UI though for now at least, mainly exists for the unique constraint. 

This will have to be revisited soon enough when municipality code gets added to the constraint. Or if we actually decide to show this somewhere as is, then this format should be smarter (maybe closer to jore3?). Not the scope of this PR though.
___
*[`test/hasura/hsl/datasets/factories.ts` line 7 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NLa5xUxDdFFopzNKp3B:-NLeFcr57BEPjmNSKlnD:b-oufacu) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/hsl/datasets/factories.ts#L7)):*

these mappers from generic to HSL are not flexible at all. In fact I the current buildLine solution is not good either. Here's a suggestion for a quick fix that should allow flexibility but won't require you to refactor the whole codebase:

```
// Make given keys of a type required, leave the rest as-is
export type RequiredKeys<T, K extends keyof T> = Required<Pick<T, K>> &
  Omit<T, K>;

export const buildHSLLine = (input: Pick<Line, 'label' | 'vehicleMode'>) => {
  const {label, vehicleMode} = input;
  return {
  name_i18n: buildLocalizedString(`transport ${vehicleMode} line ${label}`),
  short_name_i18n: buildLocalizedString(`line ${label}`),
  primary_vehicle_mode: vehicleMode,
  type_of_line: buildTypeOfLine(vehicleMode),
  transport_target: HslTransportTarget.HelsinkiInternalTraffic,
  ...input,
  } as HSLLine;
})

```

(you may need to set some attributes of the Line type as optional)

This new buildHSLLine:
1. can be used to build lines based on label and vehicle mode
2. allows the caller to set other attributes too (if e.g. want to manually override the priority for some reason)
3. can be used as a mapping function (instead of the genericLineToHsl, you could just call this buildHSLLine with the same parameters)
</blockquote></details>

I don't like how this would require us to add fields here any time we add some to generic schema. Feels troublesome, I would rather avoid that.

Also, this approach doesn't work nicely with the route methods, because the (generic) route generator does not take route properties as a parameter (unlike the Line version) but a postfix instead (based on which the route object is created).

Took some idea from your suggestion, and managed to combine the build and generic-to methods to one. Still not quite happy with the result, but maybe good enough for now.
___
*[`test/hasura/hsl/datasets/setup.ts` line 2 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NLa7rk38PpbRUD1dmyZ:-NLemIkt1qi2aD8gvPOm:b62t4h1) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/hsl/datasets/setup.ts#L2)):*

should you rename this to "hslDefaultTableConfig" for clarity?
</blockquote></details>

Renamed locally :ok_hand: 
___
*[`test/hasura/hsl/datasets/types.ts` line 12 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NLa4_Cp1Z3Ltg-TTJM_:-NLemXRXDECaatkif_wt:b85qh8q) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/hsl/datasets/types.ts#L12)):*

that's a weird looking enum key here :)
</blockquote></details>

Crap, looks like my multi-cursor-fu failed me. Good catch, fixed!
___
*[`test/hasura/hsl/datasets/types.ts` line 32 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NLa8O0hCIXAU3C18PUT:-NLergs-DlImFznr3jfI:b-nfjmz4) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/hsl/datasets/types.ts#L32)):*

this is not a type but a constant. Let's rather have it as camelCase. Same for the other Props arrays
</blockquote></details>

Goes for the generic dataset as well. Will rename all in separate commit.
___
*[`test/hasura/hsl/datasets/defaultSetup/index.ts` line 11 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NLa8a2JDkyuUP0ahU01:-NLev6_fCEIEcNFSofE3:b-6brsvs) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/hsl/datasets/defaultSetup/index.ts#L11)):*

I have to say this is a quite ugly way to override the original values :D But let's leave it like this as the whole TableConfig shit is gonna be refactored anyway to separate the data from the schema properly
</blockquote></details>

:thumbsup:  yep! Would be easier with a different kind of representation for test data.
___
*[`test/hasura/generic/datasets/setup.ts` line 47 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NL_aCxJ3CIPk4WtJf8U:-NLe3_Xg0LqbxTSfKBws:b80z1vi) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/a2667a89c3090133b0e65d419276c37fc81b667f/test/hasura/generic/datasets/setup.ts#L47)):*

if you removed this, then you could also remove it from the other `queryRunner = queryRunner.query(...)` calls
</blockquote></details>

Actually, think I'll just revert this part, and instead add `return this;` to the `insertFromJson` method. Keeps queryRunner methods consistent and fixes the typing issue why I did this change in the first place.
___
*[`test/hasura/generic/util/db.ts` line 21 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NL_bgC96inpM-iSPUlg:-NLe86uOA1bV_cTWs1vp:bc7fpad) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/generic/util/db.ts#L21)):*

this is not a very good coding pattern, it's not flexible at all. I'd rather like to see it implemented like this:
- refactoring the parameterSymbols calculation to its own function
- refactoring the value mapping to its own function

also, if it's a possibility, just take the props array directly from the TableConfig object instead of calculating the parameterSymbols. (it's available in the setupDb function where this function is called from)
</blockquote></details>

:ok_hand: split into smaller methods (nested).

Using TableConfig would be neat, but this would require some extra checks / handling for the jsonObjects (they might have different properties, or properties in inconsistent different order), so won't do that.

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132)* status: 0 of 110 files reviewed, 3 unresolved discussions (waiting on @eleksi and @Leitsi)

___
*[`test/hasura/hsl/datasets/factories.ts` line 7 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NLa5xUxDdFFopzNKp3B:-NLtdE1j8FZ9Yw6ZC5Dt:b-tpsv9w) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/hsl/datasets/factories.ts#L7)):*

I don't like how this would require us to add fields here any time we add some to generic schema. Feels troublesome, I would rather avoid that.

Also, this approach doesn't work nicely with the route methods, because the (generic) route generator does not take route properties as a parameter (unlike the Line version) but a postfix instead (based on which the route object is created).

Took some idea from your suggestion, and managed to combine the build and generic-to methods to one. Still not quite happy with the result, but maybe good enough for now.
</blockquote></details>

yeah, that would've been my step 2, to combine the generic and the HSL builders, to work with the same logic. I see you kept the generic builder as-is, but this solution looks nice too. Thanks for the refactor!
___
*[`test/hasura/generic/util/db.ts` line 21 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NL_bgC96inpM-iSPUlg:-NLtcSsz7cSbtsF6mzpg:b-84fdvq) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/generic/util/db.ts#L21)):*

:ok_hand: split into smaller methods (nested).

Using TableConfig would be neat, but this would require some extra checks / handling for the jsonObjects (they might have different properties, or properties in inconsistent different order), so won't do that.
</blockquote></details>

Inline functions... ?! well, I guess it's fine for now... it's anyway a bit of a questionmark what the scope should be as this is a class.
___
*[`test/hasura/generic/util/db.ts` line 45 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NLtcqhj5CJ16mE3RRNT:-NLtcqhj5CJ16mE3RRNU:b-7m74ox) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/aafc0a5294dda42f27c876999d7ce139896f0e46/test/hasura/generic/util/db.ts#L45)):*
> ```XML
> 
>     jsonObjects.forEach((jsonObject) => {
>       const columns = getInsertColumns(jsonObject);
> ```

nice!

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132)* status: 0 of 110 files reviewed, 3 unresolved discussions (waiting on @eleksi and @pallost)

___
*[`test/hasura/generic/util/db.ts` line 21 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NL_bgC96inpM-iSPUlg:-NLtfx-x4b-eM6z81tWm:b-njv6l2) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/1c71b5dbcc135967f43407644249f14edd5b0da1/test/hasura/generic/util/db.ts#L21)):*

Inline functions... ?! well, I guess it's fine for now... it's anyway a bit of a questionmark what the scope should be as this is a class.
</blockquote></details>

Private methods in the QueryRUnner might have been neater, but the linter doesn't allow class methods that don't reference `this`. So would've had to move these outside the class/file, which would've been even uglier and maybe harder to understand: these are quite tightly tied to the SQL query being ran here, so moving them further from that context wouldn't make much sense in my opinion.

So as I see it, we can:
- go with this
- revert to the first version, no nested functions
- something else?
- more major rework (not worth at this point)
___
*[`test/hasura/generic/util/db.ts` line 45 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132#-NLtcqhj5CJ16mE3RRNT:-NLtiO6R4FAG11te7IED:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/aafc0a5294dda42f27c876999d7ce139896f0e46/test/hasura/generic/util/db.ts#L45)):*

nice!
</blockquote></details>

Done.

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132)* status: 0 of 110 files reviewed, 1 unresolved discussion (waiting on @Leitsi)

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/132)* status: 0 of 110 files reviewed, all discussions resolved

PR #133, opened 2023-01-18
Add HSL municipality code to routes

Leitsi:
See https://github.com/HSLdevcom/jore4/issues/979 (does not resolve, also needs work on the importer).

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,15 +183,14 @@ If you wish to apply HSL migrations in local development environment:
 - now you should be able to see and apply the hsl migrations with `hasura migrate` commands
 
 When adding new metadata files, they can be added in the `metadata/hsl` directory in the correct location.
-However, when adding content to existing metadata files, a new file should be created in the
-directory of patched file, inside `/patch` directory located in the directory where the original metadata file is
-(`/metadata/hsl/databases/detault/tables/patch`). The patch file's name should match the name of the file to
-be patched. The added file should only contain the new lines to be added.
+These will be merged with the generic metadata when starting up the HSL Docker container.
+See the merge sccript `scripts/merge-metadata.sh` for details.
 
-For example, to add a new HSL specific relationship to `route_line` table, a new file should be created. The original
-file to be patched would be `/metadata/hsl/databases/detault/tables/route_line.yaml`.
+For example, to add a new HSL specific relationship to `route_line` table, a new file should be created.
+The original file to be patched would be `/metadata/hsl/databases/detault/tables/route_line.yaml`.
 
-`/metadata/hsl/databases/detault/tables/patch/route_line.yaml`:
+Patch file:
+`/metadata/hsl/databases/detault/tables/route_line.yaml`:
 
 ```
 object_relationships:
--- a/metadata/hsl/databases/default/tables/hsl_route_hsl_municipality_code.yaml
+++ b/metadata/hsl/databases/default/tables/hsl_route_hsl_municipality_code.yaml
@@ -1,0 +1,11 @@
+table:
+  name: hsl_municipality_code
+  schema: hsl_route
+array_relationships:
+  - name: routes
+    using:
+      foreign_key_constraint_on:
+        column: hsl_municipality_code
+        table:
+          name: route
+          schema: route
--- a/metadata/hsl/databases/default/tables/route_route.yaml
+++ b/metadata/hsl/databases/default/tables/route_route.yaml
@@ -1,0 +1,4 @@
+object_relationships:
+  - name: hslMunicipalityCodeByHslMunicipalityCode
+    using:
+      foreign_key_constraint_on: hsl_municipality_code
--- a/metadata/hsl/databases/default/tables/tables.yaml
+++ b/metadata/hsl/databases/default/tables/tables.yaml
@@ -1,1 +1,2 @@
+- "!include hsl_route_hsl_municipality_code.yaml"
 - "!include hsl_route_transport_target.yaml"
--- a/migrations/hsl/default/1673620142076_add_hsl_municipality_code/down.sql
+++ b/migrations/hsl/default/1673620142076_add_hsl_municipality_code/down.sql
@@ -1,0 +1,12 @@
+ALTER TABLE route.route
+  DROP COLUMN unique_label;
+
+ALTER TABLE route.route
+  ADD COLUMN unique_label text GENERATED ALWAYS AS (label || (CASE WHEN variant IS NULL THEN '' ELSE '_' || variant::text END)) STORED;
+
+COMMENT ON COLUMN route.route.unique_label IS 'Derived from label and variant. Routes are unique for each unique label for a certain direction, priority and validity period';
+
+ALTER TABLE route.route
+  DROP COLUMN hsl_municipality_code;
+
+DROP TABLE hsl_route.hsl_municipality_code CASCADE;
--- a/migrations/hsl/default/1673620142076_add_hsl_municipality_code/up.sql
+++ b/migrations/hsl/default/1673620142076_add_hsl_municipality_code/up.sql
@@ -1,0 +1,40 @@
+CREATE TABLE hsl_route.hsl_municipality_code (
+  hsl_municipality_code text PRIMARY KEY,
+  jore3_code smallint DEFAULT NULL -- nullable, so we can possibly add other municipalities in the future that are not from jore3
+);
+
+COMMENT ON TABLE
+  hsl_route.hsl_municipality_code IS
+  'Mainly used to represent a region used for route label uniqueness: labels only have to be unique when they have same municipality code.';
+INSERT INTO hsl_route.hsl_municipality_code
+  (jore3_code, hsl_municipality_code)
+  VALUES
+  -- Based on https://github.com/HSLdevcom/jore4/blob/main/wiki/jore3_route_ids.md#first-number-9
+  (0, 'legacy_not_used'), -- Not used. Some historical routes have this prefix
+  (1, 'helsinki'), -- Helsinki
+  (2, 'espoo'), -- Espoo
+  (3, 'train_or_metro'), -- Train or metro
+  (4, 'vantaa'), -- Vantaa
+  (5, 'espoon_vantaa_regional'), -- Regional routes between Vantaa and Espoo
+  (6, 'kirkkonummi_and_siuntio'), -- Kirkkonummi and Siuntio
+  (7, 'u_lines'), -- U bus routes
+  (8, 'testing_not_used'), -- for testing purposes, not used
+  (9, 'tuusula_kerava_sipoo') -- Tuusula, Kerava and Sipoo
+  ON CONFLICT (hsl_municipality_code) DO NOTHING;
+
+-- create the hsl_municipality_code column with constraints and fill it up with some default data
+ALTER TABLE route.route
+  ADD COLUMN hsl_municipality_code text NOT NULL REFERENCES hsl_route.hsl_municipality_code (hsl_municipality_code) DEFAULT 'helsinki';
+
+-- drop the default clause from hsl_municipality_code to make it mandatory to fill it in
+ALTER TABLE route.route
+  ALTER COLUMN hsl_municipality_code DROP DEFAULT;
+
+-- Update unique_label definition
+ALTER TABLE route.route
+  DROP COLUMN unique_label;
+
+ALTER TABLE route.route
+  ADD COLUMN unique_label text GENERATED ALWAYS AS (hsl_municipality_code || '_' ||label || (CASE WHEN variant IS NULL THEN '' ELSE '_' || variant::text END)) STORED;
+
+COMMENT ON COLUMN route.route.unique_label IS 'Derived from label, HSL municipality code and variant. Routes are unique for each unique label for a certain direction, priority and validity period';
--- a/migrations/hsl/default/2000000001001_R_after_migrate_route/up.sql
+++ b/migrations/hsl/default/2000000001001_R_after_migrate_route/up.sql
@@ -1,2 +1,4 @@
 ALTER TABLE route.route
   ADD CONSTRAINT route_variant_unsigned_check CHECK (variant >= 0);
+
+CREATE INDEX idx_route_hsl_municipality_code ON route.route USING btree (hsl_municipality_code);
--- a/migrations/hsl/default/2000000001002_R_after_migrate_hsl_route/down.sql
+++ b/migrations/hsl/default/2000000001002_R_after_migrate_hsl_route/down.sql
@@ -1,0 +1,2 @@
+-- we don't drop functions and constraints here, they are dropped in the before_migrate hook
+SELECT 1;
--- a/migrations/hsl/default/2000000001002_R_after_migrate_hsl_route/up.sql
+++ b/migrations/hsl/default/2000000001002_R_after_migrate_hsl_route/up.sql
@@ -1,0 +1,1 @@
+CREATE UNIQUE INDEX jore3_code_unique_idx ON hsl_route.hsl_municipality_code USING btree (jore3_code) WHERE (jore3_code IS NOT NULL);
--- a/scripts/merge-metadata.sh
+++ b/scripts/merge-metadata.sh
@@ -6,6 +6,8 @@ set -eu
 # 1. If the file with the same name already exists both in the generic and in the HSL metadata,
 # then the HSL version will be merged into the generic one
 # 2. If the file in the HSL metadata does not exist in the generic one, it will be copied
+#
+# Note: these need to be absolute paths.
 SOURCE_DIR="$1"
 DESTINATION_DIR="$2"
 
--- a/test/hasura/hsl/datasets/defaultSetup/routes.ts
+++ b/test/hasura/hsl/datasets/defaultSetup/routes.ts
@@ -1,5 +1,9 @@
 import { routes as genericRoutes } from '@datasets-generic/defaultSetup/routes';
-import { HslRoute, RouteDirection } from '@datasets-hsl/types';
+import {
+  HslMunicipalityCodeId,
+  HslRoute,
+  RouteDirection,
+} from '@datasets-hsl/types';
 import { LocalDate } from 'local-date';
 import { buildHslRoute } from '../factories';
 import { hslLines } from './lines';
@@ -12,6 +16,7 @@ export const hslRoutes: HslRoute[] = genericRoutes
         route_id: '4dadce26-0689-4843-b9fc-1ffb43a35fc1',
         on_line_id: hslLines[4].line_id,
         variant: 3,
+        hsl_municipality_code: HslMunicipalityCodeId.Helsinki,
         direction: RouteDirection.Westbound,
         priority: 20,
         validity_start: null,
--- a/test/hasura/hsl/datasets/factories.ts
+++ b/test/hasura/hsl/datasets/factories.ts
@@ -2,7 +2,12 @@ import {
   buildLine as buildGenericLine,
   buildRoute as buildGenericRoute,
 } from '@datasets-generic/factories';
-import { HslLine, HslRoute, HslTransportTarget } from '@datasets-hsl/types';
+import {
+  HslLine,
+  HslMunicipalityCodeId,
+  HslRoute,
+  HslTransportTarget,
+} from '@datasets-hsl/types';
 
 export type RequiredKeys<T, K extends keyof T> = Required<Pick<T, K>>;
 
@@ -22,6 +27,7 @@ export const buildHslRoute = (
 ): HslRoute => {
   return {
     ...(postfix ? buildGenericRoute(postfix) : {}),
+    hsl_municipality_code: HslMunicipalityCodeId.Helsinki,
     variant: null,
     ...(input || {}),
   } as HslRoute;
--- a/test/hasura/hsl/datasets/types.ts
+++ b/test/hasura/hsl/datasets/types.ts
@@ -34,8 +34,31 @@ export const hslLineProps: Property[] = [
   'transport_target',
 ];
 
+export enum HslMunicipalityCodeId {
+  LegacyNotUsed = 'legacy_not_used',
+  Helsinki = 'helsinki',
+  Espoo = 'espoo',
+  TrainOrMetro = 'train_or_metro',
+  Vantaa = 'vantaa',
+  EspoonVantaaRegional = 'espoon_vantaa_regional',
+  KirkkonummiAndSiuntio = 'kirkkonummi_and_siuntio',
+  U_Lines = 'u_lines',
+  TestingNotUsed = 'testing_not_used',
+  TuusulaKeravaSipoo = 'tuusula_kerava_sipoo',
+}
+
+export type HslMunicipalityCode = {
+  hsl_municipality_code: HslMunicipalityCodeId;
+  jore3_code: number | null;
+};
+
 export type HslRoute = GenericRoute & {
+  hsl_municipality_code: HslMunicipalityCodeId | null;
   variant: number | null;
 };
 
-export const hslRouteProps: Property[] = [...genericRouteProps, 'variant'];
+export const hslRouteProps: Property[] = [
+  ...genericRouteProps,
+  'hsl_municipality_code',
+  'variant',
+];
--- a/test/hasura/hsl/datasets/types.ts
+++ b/test/hasura/hsl/datasets/types.ts
@@ -35,7 +35,7 @@ export const hslLineProps: Property[] = [
 ];
 
 export type HslRoute = GenericRoute & {
-  variant?: number | null;
+  variant: number | null;
 };
 
 export const hslRouteProps: Property[] = [...genericRouteProps, 'variant'];
--- a/test/hasura/hsl/integration-tests/data-integrity/line-routes/hsl-municipality-code/insert-route.spec.ts
+++ b/test/hasura/hsl/integration-tests/data-integrity/line-routes/hsl-municipality-code/insert-route.spec.ts
@@ -1,0 +1,187 @@
+import * as config from '@config';
+import { hslLines } from '@datasets-hsl/defaultSetup/lines';
+import { hslRoutes } from '@datasets-hsl/defaultSetup/routes';
+import { buildHslRoute } from '@datasets-hsl/factories';
+import { getPropNameArray, queryTable, setupDb } from '@datasets-hsl/setup';
+import {
+  HslRoute,
+  RouteDirection,
+  hslRouteProps,
+  HslMunicipalityCodeId,
+} from '@datasets-hsl/types';
+import * as dataset from '@util/dataset';
+import '@util/matchers';
+import { expectErrorResponse } from '@util/response';
+import { LocalDate } from 'local-date';
+import * as pg from 'pg';
+import * as rp from 'request-promise';
+
+const toBeInserted = (
+  label: string,
+  onLineId: string | undefined,
+  hslMunicipalityCode: HslMunicipalityCodeId | null,
+): Partial<HslRoute> => {
+  return buildHslRoute(
+    {
+      on_line_id: onLineId,
+      direction: RouteDirection.Westbound,
+      hsl_municipality_code: hslMunicipalityCode,
+      variant: 3,
+      priority: 20,
+      validity_start: new LocalDate('2044-05-01'),
+      validity_end: new LocalDate('2045-04-30'),
+    },
+    label,
+  );
+};
+
+const buildMutation = (
+  label: string,
+  onLineId: string | undefined,
+  hslMunicipalityCode: HslMunicipalityCodeId | null,
+) => `
+  mutation {
+    insert_route_route(objects: ${dataset.toGraphQlObject(
+      toBeInserted(label, onLineId, hslMunicipalityCode),
+      ['direction'],
+    )}) {
+      returning {
+        ${getPropNameArray(hslRouteProps).join(',')}
+      }
+    }
+  }
+`;
+
+describe('Insert route', () => {
+  let dbConnectionPool: pg.Pool;
+
+  beforeAll(() => {
+    dbConnectionPool = new pg.Pool(config.databaseConfig);
+  });
+
+  afterAll(() => dbConnectionPool.end());
+
+  beforeEach(() => setupDb(dbConnectionPool));
+
+  const shouldReturnErrorResponse = (
+    label: string,
+    onLineId: string | undefined,
+    hslMunicipalityCode: HslMunicipalityCodeId | null,
+  ) =>
+    it('should return error response', async () => {
+      await rp
+        .post({
+          ...config.hasuraRequestTemplate,
+          body: { query: buildMutation(label, onLineId, hslMunicipalityCode) },
+        })
+        .then(expectErrorResponse());
+    });
+
+  const shouldNotModifyDatabase = (
+    label: string,
+    onLineId: string | undefined,
+    hslMunicipalityCode: HslMunicipalityCodeId | null,
+  ) =>
+    it('should not modify the database', async () => {
+      await rp.post({
+        ...config.hasuraRequestTemplate,
+        body: { query: buildMutation(label, onLineId, hslMunicipalityCode) },
+      });
+
+      const response = await queryTable(dbConnectionPool, 'route.route');
+
+      expect(response.rowCount).toEqual(hslRoutes.length);
+      expect(response.rows).toEqual(expect.arrayContaining(hslRoutes));
+    });
+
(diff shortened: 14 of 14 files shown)


Leitsi:
*[`metadata/hsl/databases/default/tables/hsl_route_hsl_municipality_code.yaml` line 3 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NM2T7CA-l_kbZRm8p46:-NM2T7CA-l_kbZRm8p47:b-b92rdn) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/e04825cc7a773bb590f66db6b40dc4cf2d09016c/metadata/hsl/databases/default/tables/hsl_route_hsl_municipality_code.yaml#L3)):*
Possible problem here that I just realised: this is currently not being configured as a Hasura enum table.
Missing `is_enum` flag, compare to eg. https://github.com/HSLdevcom/jore4-hasura/tree/main/metadata/hsl/databases/default/tables/hsl_route_transport_target.yaml

Don't think we can make it into one without changing the schema, see https://hasura.io/docs/latest/schema/postgres/enums/#pg-create-enum-table :
> 1. The table must have a single-column primary key
> 3. The table must not contain any other columns

So, the `jore3_code` would have to be dropped.

Should we:
- ignore this, don't mark the table as a Hasura enum table  (I'm not sure what this would cause in practice)
- adjust the schema: drop the ´jore3_code` altogether (shouldn't be necessary to jore4, but some systems that we integrate to might want this?)
- adjust the schema: add another table for `jore3_code`, foreign key to `hsl_municipality_code` (a bit more complex but not too bad maybe)

Leitsi:
*[`metadata/hsl/databases/default/tables/hsl_route_hsl_municipality_code.yaml` line 3 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NM2T7CA-l_kbZRm8p46:-NM2UvED3xz1VgJf1NEQ:bk5nes1) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/e04825cc7a773bb590f66db6b40dc4cf2d09016c/metadata/hsl/databases/default/tables/hsl_route_hsl_municipality_code.yaml#L3)):*

Possible problem here that I just realised: this is currently not being configured as a Hasura enum table.
Missing `is_enum` flag, compare to eg. https://github.com/HSLdevcom/jore4-hasura/tree/main/metadata/hsl/databases/default/tables/hsl_route_transport_target.yaml

Don't think we can make it into one without changing the schema, see https://hasura.io/docs/latest/schema/postgres/enums/#pg-create-enum-table :
> 1. The table must have a single-column primary key
> 3. The table must not contain any other columns

So, the `jore3_code` would have to be dropped.

Should we:
- ignore this, don't mark the table as a Hasura enum table  (I'm not sure what this would cause in practice)
- adjust the schema: drop the ´jore3_code` altogether (shouldn't be necessary to jore4, but some systems that we integrate to might want this?)
- adjust the schema: add another table for `jore3_code`, foreign key to `hsl_municipality_code` (a bit more complex but not too bad maybe)
</blockquote></details>

If we go with the last option, that can also be done later if we do notice that we need the jore3 code.

culka (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133)* status: 0 of 14 files reviewed, 1 unresolved discussion (waiting on @Leitsi)

___
*[`metadata/hsl/databases/default/tables/hsl_route_hsl_municipality_code.yaml` line 3 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NM2T7CA-l_kbZRm8p46:-NM3BPBOAml9oQj9lBtr:bv0i26z) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/e04825cc7a773bb590f66db6b40dc4cf2d09016c/metadata/hsl/databases/default/tables/hsl_route_hsl_municipality_code.yaml#L3)):*

If we go with the last option, that can also be done later if we do notice that we need the jore3 code.
</blockquote></details>

If I understand the documentation, the only result of not marking the table as a Hasura enum table would be illegal values resulting in a constraint error instead of a more descriptive enum one.

It's still unsure if we will need `jore3_code` value or not, the need should clarify during the spring. Option 3 sounds good, we can either add the table later or add the table now and drop it later without affecting anything else.

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133)* status: 0 of 23 files reviewed, 1 unresolved discussion (waiting on @culka)

___
*[`metadata/hsl/databases/default/tables/hsl_route_hsl_municipality_code.yaml` line 3 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NM2T7CA-l_kbZRm8p46:-NMhipAxCZ1WtcwShf5S:b-a2lxa9) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/e04825cc7a773bb590f66db6b40dc4cf2d09016c/metadata/hsl/databases/default/tables/hsl_route_hsl_municipality_code.yaml#L3)):*

If I understand the documentation, the only result of not marking the table as a Hasura enum table would be illegal values resulting in a constraint error instead of a more descriptive enum one.

It's still unsure if we will need `jore3_code` value or not, the need should clarify during the spring. Option 3 sounds good, we can either add the table later or add the table now and drop it later without affecting anything else.
</blockquote></details>

After clarifying the requirements from HSL, decided to leave this as is:
- the whole municipality code is legacy and should be deleted eventually (= maybe not worth to implement anything too complex for it)
- we already know that we need the jore3 code (actually it is the main thing, we don't really care what municipality it really means I think) -> add it already

Removed the unique_label adjustments though: municipality shouldn't be considered in those after all.

pallost (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133)* status: 0 of 21 files reviewed, 9 unresolved discussions (waiting on @culka and @Leitsi)

___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 2 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5xsA6Dy-V2YkIbKaB:-NN5xsA6Dy-V2YkIbKaC:b-dxt92j) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L2)):*
> ```TSQL
> CREATE TABLE hsl_route.legacy_hsl_municipality_code (
>   legacy_hsl_municipality_code text PRIMARY KEY,
> ```

I think it's the numeric value that's legacy, the text value is current. I'd just name the columns as "municipality_code" and "jore3_code" as the table's name already tells that this is legacy
___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 3 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5vsLB0dbWR6dEzuS4:-NN5vsLB0dbWR6dEzuS5:b-dgik26) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L3)):*
> ```TSQL
> CREATE TABLE hsl_route.legacy_hsl_municipality_code (
>   legacy_hsl_municipality_code text PRIMARY KEY,
>   jore3_code smallint NOT NULL CHECK (0 <= jore3_code AND jore3_code <= 9) -- all possible values are inserted here.
> ```

check constraints should be implemented as repeatable migrations, let's move this to /migrations/hsl/default/2000000001001_R_after_migrate_route

(note: you may need to tweak the _R_before_migrate script to also wipe the "hsl_route" schema)
___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 27 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5zIFYE72X-L_Z_Gbd:-NN5zIFYE72X-L_Z_Gbe:b-fj6jbf) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L27)):*
> ```TSQL
> -- Create the legacy_hsl_municipality_code column with constraints and fill it up with some default data.
> -- Currently all routes in Jore3 have this so NOT NULL is okay.
> -- Should probably be made nullable once there is a new mechanism in place to replace this.
> ```

if the municipality code is only used when importing from jore3, then I'd rather just make the column NULLABLE (and no default values either). If it's a critical column when exporting some data from jore4, then it has to be NOT NULLABLE.
___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 29 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5xeWW5xdCv3HxwsXF:-NN5xeWW5xdCv3HxwsXG:b-8e29lk) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L29)):*
> ```TSQL
> -- Should probably be made nullable once there is a new mechanism in place to replace this.
> ALTER TABLE route.route
>   ADD COLUMN legacy_hsl_municipality_code text NOT NULL REFERENCES hsl_route.legacy_hsl_municipality_code (legacy_hsl_municipality_code) DEFAULT 'helsinki';
> ```

I'm wondering, what's carrying more value for us from the municipality codes
___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 35 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN6-DXv122822PVsolM:-NN6-DXv122822PVsolN:b-y0bqiy) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L35)):*
> ```TSQL
>   ALTER COLUMN legacy_hsl_municipality_code DROP DEFAULT;
> 
> CREATE INDEX idx_route_legacy_hsl_municipality_code ON route.route USING btree (legacy_hsl_municipality_code);
> ```

nice, thanks for the index!
___
*[`migrations/hsl/default/1674218837216_add_line_legacy_hsl_municipality_code/up.sql` line 5 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5zqp88uHjbnUWyPw_:-NN5zqp88uHjbnUWyPwa:bs4ptxe) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1674218837216_add_line_legacy_hsl_municipality_code/up.sql#L5)):*
> ```TSQL
> -- Should probably be made nullable once there is a new mechanism in place to replace this.
> ALTER TABLE route.line
>   ADD COLUMN legacy_hsl_municipality_code text NOT NULL REFERENCES hsl_route.legacy_hsl_municipality_code (legacy_hsl_municipality_code) DEFAULT 'helsinki';
> ```

same as above, for route.route. Should think whether this is nullable or not
___
*[`test/hasura/hsl/datasets/types.ts` line 39 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN608me6PuEEpzBz0Cg:-NN608me6PuEEpzBz0Ch:b-w5z2g) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/test/hasura/hsl/datasets/types.ts#L39)):*
> ```XML
> ];
> 
> export enum LegacyHslMunicipalityCodeId {
> ```

Is it "Code" or "Id"? Maybe not "CodeId" :)
___
*[`test/hasura/hsl/datasets/types.ts` line 47 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN602dv1p0fmsoX7Hlu:-NN602dv1p0fmsoX7Hlv:b-2xoscm) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/test/hasura/hsl/datasets/types.ts#L47)):*
> ```XML
>   EspoonVantaaRegional = 'espoon_vantaa_regional',
>   KirkkonummiAndSiuntio = 'kirkkonummi_and_siuntio',
>   U_Lines = 'u_lines',
> ```

`ULines` instead of `U_Lines`?
___
*[`metadata/hsl/databases/default/tables/hsl_route_hsl_municipality_code.yaml` line 3 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NM2T7CA-l_kbZRm8p46:-NN60_AEBYi6JZlEWe4F:b-susrbz) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/e04825cc7a773bb590f66db6b40dc4cf2d09016c/metadata/hsl/databases/default/tables/hsl_route_hsl_municipality_code.yaml#L3)):*

After clarifying the requirements from HSL, decided to leave this as is:
- the whole municipality code is legacy and should be deleted eventually (= maybe not worth to implement anything too complex for it)
- we already know that we need the jore3 code (actually it is the main thing, we don't really care what municipality it really means I think) -> add it already

Removed the unique_label adjustments though: municipality shouldn't be considered in those after all.
</blockquote></details>

yeah, this is rather a "mapping" table instead of a plain "enum" table. This is good as-is IMHO

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133)* status: 0 of 21 files reviewed, 8 unresolved discussions (waiting on @culka and @pallost)

___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 2 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5xsA6Dy-V2YkIbKaB:-NN6hBGxEgtbT1UxpvoV:b-3voujq) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L2)):*

I think it's the numeric value that's legacy, the text value is current. I'd just name the columns as "municipality_code" and "jore3_code" as the table's name already tells that this is legacy
</blockquote></details>

True, changing.
___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 3 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5vsLB0dbWR6dEzuS4:-NN6gAmoEK7da3-HREkJ:b-guj088) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L3)):*

check constraints should be implemented as repeatable migrations, let's move this to /migrations/hsl/default/2000000001001_R_after_migrate_route

(note: you may need to tweak the _R_before_migrate script to also wipe the "hsl_route" schema)
</blockquote></details>

Good point, this doesn't work (even if it would be otherwise "fine") because _all_ the constraints get dropped by repeatable migrations. Will fix.
___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 27 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5zIFYE72X-L_Z_Gbd:-NN6_F_u1UOTNGV8naJo:b-z5kh8t) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L27)):*
> If it's a critical column when exporting some data from jore4, then it has to be NOT NULLABLE.

As I understood, this is the case currently. But at some point we should be getting rid of this, = drop the whole municipality or replace with some other entity, and from then on routes/lines would not need this = THEN we could make this nullable. But until then, better to ensure this is always set.
___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 29 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5xeWW5xdCv3HxwsXF:-NN6hD9x-wjhp7Ai_h-H:b9m4eds) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L29)):*

I'm wondering, what's carrying more value for us from the municipality codes
</blockquote></details>

I'm not quite grasping what you're trying to say here.
___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 35 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN6-DXv122822PVsolM:-NN6_DaJ1d4f994Ql5JO:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L35)):*

nice, thanks for the index!
</blockquote></details>

Done.
___
*[`migrations/hsl/default/1674218837216_add_line_legacy_hsl_municipality_code/up.sql` line 5 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5zqp88uHjbnUWyPw_:-NN6a2Es7IVn3DtLe_bM:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1674218837216_add_line_legacy_hsl_municipality_code/up.sql#L5)):*

same as above, for route.route. Should think whether this is nullable or not
</blockquote></details>

Done.
___
*[`test/hasura/hsl/datasets/types.ts` line 39 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN608me6PuEEpzBz0Cg:-NN6bmauCLTYdrARto-A:b-gvby72) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/test/hasura/hsl/datasets/types.ts#L39)):*

Is it "Code" or "Id"? Maybe not "CodeId" :)
</blockquote></details>

Yeah, crappy name. The name `LegacyHslMunicipalityCode` was already taken by the table definition, just below this enum, so added the "id" here (since this enum acts as the primary key). On a second thought, will drop the suffix from here altogether, so just `LegacyHslMunicipality`
___
*[`test/hasura/hsl/datasets/types.ts` line 47 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN602dv1p0fmsoX7Hlu:-NN6aYWk8AwT6hHh0TPt:b4c9lyr) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/test/hasura/hsl/datasets/types.ts#L47)):*

`ULines` instead of `U_Lines`?
</blockquote></details>

:thumbsup:

jarkkoka (commented):
> Reviewed 7 of 14 files at r1, 7 of 21 files at r2.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133)* status: 6 of 22 files reviewed, 11 unresolved discussions (waiting on @culka, @Leitsi, and @pallost)

___
*[`migrations/generic/default/1000000000000_R_before_migrate/up.sql` line 45 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NNB2p6M3Rc9xO4pzu6F:-NNB2p6M3Rc9xO4pzu6G:b-zbuy7y) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/09def5f7f429f73992e159c2c45b3ca66b60ec65/migrations/generic/default/1000000000000_R_before_migrate/up.sql#L45)):*
> ```TSQL
>         'service_pattern',
>         -- HSL specific schemas.
>         'hsl_route'
> ```

"HSL" leaked to the general scheme. It seems that the custom repeatable migrations scheme does not currently provide a clean separation in this regard.



_Code quote:_
```TSQL
        -- HSL specific schemas.
        'hsl_route'
```
___
*[`migrations/generic/default/1000000000000_R_before_migrate/up.sql` line 67 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NNB3SST1G7WAxEW8tJZ:-NNB3SST1G7WAxEW8tJ_:b-wisdrw) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/09def5f7f429f73992e159c2c45b3ca66b60ec65/migrations/generic/default/1000000000000_R_before_migrate/up.sql#L67)):*
> ```TSQL
>         'service_pattern',
>         'timing_pattern',
>         -- HSL specific schemas.
> ```

same remark as above (leak)



_Code quote:_
```TSQL
        -- HSL specific schemas.
        'hsl_route'
```
___
*[`migrations/generic/default/1000000000000_R_before_migrate/up.sql` line 108 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NNB3_GZA8JKE8iHwIM1:-NNB3_GZA8JKE8iHwIM2:b-wisdrw) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/09def5f7f429f73992e159c2c45b3ca66b60ec65/migrations/generic/default/1000000000000_R_before_migrate/up.sql#L108)):*
> ```TSQL
>       -- HSL specific schemas.
>       'hsl_route'
>     ))
> ```

same remark as above (leak)



_Code quote:_
```TSQL
      -- HSL specific schemas.
      'hsl_route'
    ))
```

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133)* status: 6 of 22 files reviewed, 11 unresolved discussions (waiting on @culka, @Leitsi, and @pallost)

___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 2 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5xsA6Dy-V2YkIbKaB:-NNB4DnS6YIK_ZAcZXQL:b25y2c1) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L2)):*

True, changing.
</blockquote></details>

I would put the numeric column as the primary key because it must not contain duplicates. The text column is just a human friendly description field IMO.

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133)* status: 6 of 22 files reviewed, 11 unresolved discussions (waiting on @culka, @Leitsi, and @pallost)

___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 29 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5xeWW5xdCv3HxwsXF:-NNB7IIQ52DsbbbsWBjW:bnfejlb) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L29)):*

I'm not quite grasping what you're trying to say here.
</blockquote></details>

I think what Tamas meant. Generating the long label should be done solely based on the values in the `route.route` table, without looking up the numeric code from another table via SQL join.

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133)* status: 6 of 22 files reviewed, 11 unresolved discussions (waiting on @culka, @Leitsi, and @pallost)

___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 29 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5xeWW5xdCv3HxwsXF:-NNG7JTG6ooVdfsXqpSu:b-xi2io5) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L29)):*

I think what Tamas meant. Generating the long label should be done solely based on the values in the `route.route` table, without looking up the numeric code from another table via SQL join.
</blockquote></details>

This prevents using computed field for constructing "long label" for exports. Thus, constructing long labels for routes (prefixed with HSL municipality code) requires an SQL join. Tamas, how do you see this? We discussed with Tommi that constructing long labels could be done in the business logic code of export services...

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133)* status: 6 of 22 files reviewed, 11 unresolved discussions (waiting on @culka, @jarkkoka, and @pallost)

___
*[`migrations/generic/default/1000000000000_R_before_migrate/up.sql` line 45 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NNB2p6M3Rc9xO4pzu6F:-NNGMgUJ0wX4m9wZWIo-:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/09def5f7f429f73992e159c2c45b3ca66b60ec65/migrations/generic/default/1000000000000_R_before_migrate/up.sql#L45)):*

"HSL" leaked to the general scheme. It seems that the custom repeatable migrations scheme does not currently provide a clean separation in this regard.
</blockquote></details>

Done.
___
*[`migrations/generic/default/1000000000000_R_before_migrate/up.sql` line 67 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NNB3SST1G7WAxEW8tJZ:-NNGMgxl3RtyfdHyBh0K:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/09def5f7f429f73992e159c2c45b3ca66b60ec65/migrations/generic/default/1000000000000_R_before_migrate/up.sql#L67)):*

same remark as above (leak)
</blockquote></details>

Done.
___
*[`migrations/generic/default/1000000000000_R_before_migrate/up.sql` line 108 at r3](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NNB3_GZA8JKE8iHwIM1:-NNGMijJ9WDZeloMA5OQ:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/09def5f7f429f73992e159c2c45b3ca66b60ec65/migrations/generic/default/1000000000000_R_before_migrate/up.sql#L108)):*

same remark as above (leak)
</blockquote></details>

Done.
___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 2 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5xsA6Dy-V2YkIbKaB:-NNGMkR73oAUWuSVXLZO:b-x0pzfw) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L2)):*
> I would put the numeric column as the primary key because it must not contain duplicates. The text column is just a human friendly description field IMO.

Discussed this in person. Will not do, too much effort, not worth it, and this approach is acceptable as well.
___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 29 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5xeWW5xdCv3HxwsXF:-NNGNBHg2rHvFvx9ADN7:b-eddds3) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L29)):*

This prevents using computed field for constructing "long label" for exports. Thus, constructing long labels for routes (prefixed with HSL municipality code) requires an SQL join. Tamas, how do you see this? We discussed with Tommi that constructing long labels could be done in the business logic code of export services...
</blockquote></details>

Discussed this in person (with Jarkko). Same conclusion as above, not worth changing.

Alternative (in my opinion) would be to just ditch the whole enum table and add the municipality code to routes/lines merely as a magic number (the enum would not really be used anyway if we switch the primary key as proposed), but magic numbers like that hurt my soul. And there's no guarantee that we won't need to add some special logic to exports for some municipality codes, in which case enums are neater to have.

Also, the long labels (including municipality code, that is) will only be used for exports, so not worth handling them in DB level in my opinion. Implementing such computed field would not be a simple legacy_hsl_municipality_code+label+variant, but involves padding with zeros between the parts etc. So that would complicate the computed field. Would be much simpler to handle in export code, I think.

pallost (approved):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133)* status: 6 of 27 files reviewed, 4 unresolved discussions (waiting on @culka and @jarkkoka)

___
*[`migrations/generic/default/1000000000000_R_before_migrate/up.sql` line 24 at r4](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NNGRN0FEMMsw9-PHiTN:-NNGRN0G2k5NI3Ib_zix:b6iys8g) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/c6b149dc0fe11cf6ab172914607a4e12ac49bad6/migrations/generic/default/1000000000000_R_before_migrate/up.sql#L24)):*
> ```TSQL
> -- note2: pg_catalog.pg_triggers contains the TRUNCATE triggers but many other things too
> -- so here we combine a bit of both
> CREATE OR REPLACE FUNCTION drop_triggers(target_schemas text[]) RETURNS void
> ```

good idea to wrap this into a function!
___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 27 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5zIFYE72X-L_Z_Gbd:-NNGRxFzErntciKopJpg:b-5eav66) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L27)):*

> If it's a critical column when exporting some data from jore4, then it has to be NOT NULLABLE.

As I understood, this is the case currently. But at some point we should be getting rid of this, = drop the whole municipality or replace with some other entity, and from then on routes/lines would not need this = THEN we could make this nullable. But until then, better to ensure this is always set.
</blockquote></details>

UI devs will hate you ;)
___
*[`migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql` line 29 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/133#-NN5xeWW5xdCv3HxwsXF:-NNGS9cl5VJWO2nKKbt-:bkxoofp) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3e9cdf6f71759c5dec10294c3488dad3d64bce9e/migrations/hsl/default/1673620142076_add_route_legacy_hsl_municipality_code/up.sql#L29)):*

Discussed this in person (with Jarkko). Same conclusion as above, not worth changing.

Alternative (in my opinion) would be to just ditch the whole enum table and add the municipality code to routes/lines merely as a magic number (the enum would not really be used anyway if we switch the primary key as proposed), but magic numbers like that hurt my soul. And there's no guarantee that we won't need to add some special logic to exports for some municipality codes, in which case enums are neater to have.

Also, the long labels (including municipality code, that is) will only be used for exports, so not worth handling them in DB level in my opinion. Implementing such computed field would not be a simple legacy_hsl_municipality_code+label+variant, but involves padding with zeros between the parts etc. So that would complicate the computed field. Would be much simpler to handle in export code, I think.
</blockquote></details>

apparently my comment got lost when got formatted :) Yeah, let's just go with this...

PR #140, opened 2023-02-02
Separate seed data and table schema

pallost:
This enables easier separation between generic and hsl schema, also simplifies modifying the default dataset for special tests' needs

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/140)

### DIFF
--- a/scripts/prettify.sh
+++ b/scripts/prettify.sh
@@ -2,4 +2,4 @@
 
 set -eu
 
-npx prettier --write .
+npx prettier@2.4.1 --write .
--- a/test/hasura/generic/datasets/defaultSetup/index.ts
+++ b/test/hasura/generic/datasets/defaultSetup/index.ts
@@ -1,14 +1,4 @@
-import {
-  infrastructureLinkAlongRouteProps,
-  infrastructureLinkProps,
-  lineProps,
-  routeProps,
-  scheduledStopPointInvariantProps,
-  scheduledStopPointProps,
-  timingPatternTimingPlaceProps,
-  vehicleModeOnScheduledStopPointProps,
-  vehicleSubmodeOnInfrastructureLinkProps,
-} from '@datasets-generic/types';
+import { GenericNetworkDbTables } from '@datasets-generic/schema';
 import {
   infrastructureLinks,
   vehicleSubmodeOnInfrastructureLink,
@@ -22,50 +12,42 @@ import {
 } from './scheduled-stop-points';
 import { timingPlaces } from './timing-places';
 
-export const defaultTableConfig: TableLikeConfig[] = [
-  {
-    name: 'infrastructure_network.infrastructure_link',
-    data: infrastructureLinks,
-    props: infrastructureLinkProps,
-  },
-  {
-    name: 'infrastructure_network.vehicle_submode_on_infrastructure_link',
-    data: vehicleSubmodeOnInfrastructureLink,
-    props: vehicleSubmodeOnInfrastructureLinkProps,
-  },
-  {
-    name: 'timing_pattern.timing_place',
-    data: timingPlaces,
-    props: timingPatternTimingPlaceProps,
-  },
-  {
-    name: 'service_pattern.scheduled_stop_point_invariant',
-    data: scheduledStopPointInvariants,
-    props: scheduledStopPointInvariantProps,
-  },
-  {
-    name: 'service_pattern.scheduled_stop_point',
-    data: scheduledStopPoints,
-    props: scheduledStopPointProps,
-  },
-  {
-    name: 'service_pattern.vehicle_mode_on_scheduled_stop_point',
-    data: vehicleModeOnScheduledStopPoint,
-    props: vehicleModeOnScheduledStopPointProps,
-  },
-  { name: 'route.line', data: lines, props: lineProps },
-  { name: 'route.route', data: routes, props: routeProps },
-  {
-    name: 'route.infrastructure_link_along_route',
-    data: infrastructureLinkAlongRoute,
-    props: infrastructureLinkAlongRouteProps,
-  },
-
-  // views
-  {
-    name: 'service_pattern.scheduled_stop_point',
-    props: scheduledStopPointProps,
-    isView: true,
-  },
-  { name: 'route.route', props: routeProps, isView: true },
-];
+export const defaultGenericNetworkDbData: TableData<GenericNetworkDbTables>[] =
+  [
+    {
+      name: 'infrastructure_network.infrastructure_link',
+      data: infrastructureLinks,
+    },
+    {
+      name: 'infrastructure_network.vehicle_submode_on_infrastructure_link',
+      data: vehicleSubmodeOnInfrastructureLink,
+    },
+    {
+      name: 'timing_pattern.timing_place',
+      data: timingPlaces,
+    },
+    {
+      name: 'service_pattern.scheduled_stop_point_invariant',
+      data: scheduledStopPointInvariants,
+    },
+    {
+      name: 'service_pattern.scheduled_stop_point',
+      data: scheduledStopPoints,
+    },
+    {
+      name: 'service_pattern.vehicle_mode_on_scheduled_stop_point',
+      data: vehicleModeOnScheduledStopPoint,
+    },
+    {
+      name: 'route.line',
+      data: lines,
+    },
+    {
+      name: 'route.route',
+      data: routes,
+    },
+    {
+      name: 'route.infrastructure_link_along_route',
+      data: infrastructureLinkAlongRoute,
+    },
+  ];
--- a/test/hasura/generic/datasets/prioritizedRouteVerification/index.ts
+++ b/test/hasura/generic/datasets/prioritizedRouteVerification/index.ts
@@ -4,19 +4,7 @@ import {
   scheduledStopPointInJourneyPattern,
   scheduledStopPointInJourneyPatternWithTempRoute,
 } from '@datasets-generic/prioritizedRouteVerification/journey-patterns';
-import {
-  infrastructureLinkAlongRouteProps,
-  infrastructureLinkProps,
-  journeyPatternProps,
-  lineProps,
-  routeProps,
-  scheduledStopPointInJourneyPatternProps,
-  scheduledStopPointInvariantProps,
-  scheduledStopPointProps,
-  timingPatternTimingPlaceProps,
-  vehicleModeOnScheduledStopPointProps,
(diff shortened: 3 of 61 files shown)


Leitsi (changes requested):
> Reviewed 60 of 61 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/140)* status: all files reviewed, 5 unresolved discussions (waiting on @pallost)

___
*[`scripts/prettify.sh` line 5 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/140#-NNH5jk0ElhcXf4DeoLb:-NNH5jk14PnTtYVobkAy:b-slan6t) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/f451e8c64cf87b5fabb457e854dc48d3f0856a22/scripts/prettify.sh#L5)):*
> ```Shell
> set -eu
> 
> npx prettier@2.4.1 --write .
> ```

Not great, since now the version info needs to be updated in two places. But don't really have any silver bullet to offer here. Moving the prettier install to a new package.json on project root just for prettier should work, but is it worth the trouble?
___
*[`test/hasura/generic/datasets/schema.ts` line 30 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/140#-NNHQN8899oPZNtahJRk:-NNHQN8899oPZNtahJRl:b-cft45u) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3014a9af790874a9c481f114255cea2a0d446fa1/test/hasura/generic/datasets/schema.ts#L30)):*
> ```XML
> export type GenericNetworkDbTables = typeof genericNetworkDbTables[number];
> 
> export const genericNetworkDbSchema: TableSchema<GenericNetworkDbTables>[] = [
> ```

Hmm, seeing how this is pretty much exclusively used with `findTableSchema`, feels a bit difficult to have this as an array instead of `Record<TableName, Property[]>` object. Of course, with this it is more symmetric with the `TableData` objects, though I'm not sure if that provides any value. But, we did talk about this structure before this got implemented, so I suppose this is fine.
___
*[`test/hasura/generic/datasets/prioritizedRouteVerification/index.ts` line 95 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/140#-NNHOBCS3IWZQucHFXoJ:-NNHOBCS3IWZQucHFXoK:b6e06qs) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/6029ef3d1c6678fd14c8e4e2296f2d83ce066db1/test/hasura/generic/datasets/prioritizedRouteVerification/index.ts#L95)):*
> ```XML
>     isView: true,
>   },
>   { name: 'route.route', props: routeProps, isView: true },
> ```

This removing of isView would've been nice to do in its own commit to more clearly show it really was unnecessary, but not worth splitting anymore indeed. :heavy_check_mark: 
___
*[`test/hasura/hsl/schema.ts` line 8 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/140#-NNHTks8A7z6MSSaYVh7:-NNHTks9AyvSphFNrple:b-soi85k) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3014a9af790874a9c481f114255cea2a0d446fa1/test/hasura/hsl/schema.ts#L8)):*
> ```XML
> import { upsertList } from '@util/schema';
> 
> export const hslNetworkDbTables = [
> ```

Should hsl and generic folders have similar structure? Mostly they do, but now the schema files are located in different places (generic is under datasets and hsl is not):

```
test/hasura/generic/datasets/schema.ts
test/hasura/hsl/schema.ts
```
___
*[`test/hasura/util/schema.ts` line 17 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/140#-NNHUScrBOgPw-C28xvJ:-NNHUScrBOgPw-C28xvK:bkmpyh3) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3014a9af790874a9c481f114255cea2a0d446fa1/test/hasura/util/schema.ts#L17)):*
> ```XML
> 
> export type ValueFunction<TItem> = (item: TItem) => ExplicitAny;
> export const upsertList = <TItem extends ExplicitAny>(
> ```

The name `upsertList` sounds like the list itself is updated/inserted, when actually its the list items that do. How about eg. `mergeLists` instead, since we have two lists that we are combining here (lodash merge for objects works in a similar way, values in 2nd list override those in 1st one).
___
*[`test/hasura/util/schema.ts` line 26 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/140#-NNHY2IbBA1-Vt404R34:-NNHY2IbBA1-Vt404R35:b-3fd6sx) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3014a9af790874a9c481f114255cea2a0d446fa1/test/hasura/util/schema.ts#L26)):*
> ```XML
> 
>   // these items do exist in the original list, they are replaced
>   const replacedItems = originalItems.reduce<TItem[]>((sum, currentItem) => {
> ```

`sum` is confusing here, we are not dealing with numbers. `result`?

pallost (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/140)* status: 54 of 61 files reviewed, 5 unresolved discussions (waiting on @Leitsi)

___
*[`scripts/prettify.sh` line 5 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/140#-NNH5jk0ElhcXf4DeoLb:-NNLNbsPEfhRMynnLgKc:b-2ace7b) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/f451e8c64cf87b5fabb457e854dc48d3f0856a22/scripts/prettify.sh#L5)):*

Not great, since now the version info needs to be updated in two places. But don't really have any silver bullet to offer here. Moving the prettier install to a new package.json on project root just for prettier should work, but is it worth the trouble?
</blockquote></details>

You are right... but ehh, I was playing around with this a little, but couldn't find a nice solution: 
- the root folder has different prettier configurations compared to the test folder
- the test folder also has eslint-prettier baked together, so would need to solve that the test folder also sees the root folder's prettier installation somehow
- all in all, I would need to introduce yarn workspaces probably, but that's probably not worth the hassle only for this.

I was wondering actually that we were anyway considering creating own hasura instances for networkdb and timetablesdb. These would however be smart to be just placed to the same repo for simplicity... but this would anyway require yarn workspaces to be used for the tests, so that would be an ideal moment to fix this prettier config too. But let's return to this later :)
___
*[`test/hasura/generic/datasets/schema.ts` line 30 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/140#-NNHQN8899oPZNtahJRk:-NNLOPpD5PNhOLH9cZa1:bcq5d0p) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3014a9af790874a9c481f114255cea2a0d446fa1/test/hasura/generic/datasets/schema.ts#L30)):*

Hmm, seeing how this is pretty much exclusively used with `findTableSchema`, feels a bit difficult to have this as an array instead of `Record<TableName, Property[]>` object. Of course, with this it is more symmetric with the `TableData` objects, though I'm not sure if that provides any value. But, we did talk about this structure before this got implemented, so I suppose this is fine.
</blockquote></details>

I was actually wondering the same, maybe a hash-table would be a better approach. I'd however make it slightly more flexible... I'll show you how in a fixup commit
___
*[`test/hasura/generic/datasets/prioritizedRouteVerification/index.ts` line 95 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/140#-NNHOBCS3IWZQucHFXoJ:-NNLOdit8HjzspzBj6Es:bd7xk5f) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/6029ef3d1c6678fd14c8e4e2296f2d83ce066db1/test/hasura/generic/datasets/prioritizedRouteVerification/index.ts#L95)):*

This removing of isView would've been nice to do in its own commit to more clearly show it really was unnecessary, but not worth splitting anymore indeed. :heavy_check_mark: 
</blockquote></details>

I'm sorry, it was pretty hard to fix a single thing at once. I approached the whole thing by just redefining the types, so that kinda dragged the whole refactoring with itself :)
___
*[`test/hasura/util/schema.ts` line 17 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/140#-NNHUScrBOgPw-C28xvJ:-NNLOxHA9R_W1a6SCOZ0:b-pnrmh) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3014a9af790874a9c481f114255cea2a0d446fa1/test/hasura/util/schema.ts#L17)):*

The name `upsertList` sounds like the list itself is updated/inserted, when actually its the list items that do. How about eg. `mergeLists` instead, since we have two lists that we are combining here (lodash merge for objects works in a similar way, values in 2nd list override those in 1st one).
</blockquote></details>

yeah, that's a better name, thanks! Too bad lodash doesn't really implement this functionality itself...
___
*[`test/hasura/util/schema.ts` line 26 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/140#-NNHY2IbBA1-Vt404R34:-NNLPRj78gQj6KzWmwrU:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3014a9af790874a9c481f114255cea2a0d446fa1/test/hasura/util/schema.ts#L26)):*

`sum` is confusing here, we are not dealing with numbers. `result`?
</blockquote></details>

Done.
___
*[`test/hasura/hsl/schema.ts` line 8 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/140#-NNHTks8A7z6MSSaYVh7:-NNLNTlIA2qeqCEMPixE:b-suf531) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/3014a9af790874a9c481f114255cea2a0d446fa1/test/hasura/hsl/schema.ts#L8)):*

Should hsl and generic folders have similar structure? Mostly they do, but now the schema files are located in different places (generic is under datasets and hsl is not):

```
test/hasura/generic/datasets/schema.ts
test/hasura/hsl/schema.ts
```
</blockquote></details>

good catch! I accidentally put it to the wrong folder. Fixed.

Leitsi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/140#-:-NNM-xe52BcZTy_ImeIp:bnfp4nl)* status: 13 of 61 files reviewed, all discussions resolved (waiting on @pallost)

PR #144, opened 2023-02-07
Make legacyHslMunicipalityCode nullable

Leitsi:
Because of course if it is NOT NULL we couldn't create new routes/lines from Jore 4 UI until it can set the field. Which it can't currently, and not sure yet if it even should.
So, make this nullable until we specify how this should be handled for new routes/lines.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/144)


tk-vincit (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/144)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Leitsi)

PR #178, opened 2023-04-25
Add timetable version functions

HenrikHartiala:
#### Add timetable version return value table 
* This introduces the schema return_value which is used to have all the
tables that will not contain any data, but are used for function return
values as they are required to have. There are already this type of
return value tables in the database that are not located in this scheme,
but they need to be moved under this schema in the future. Although
this commit will add timetable_version as the first return_value.

#### Add get timetable version functions
* This function set is used to get timetable versions and their
effectiveness during a given time range and observation date.


Leitsi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 0 of 10 files reviewed, 11 unresolved discussions (waiting on @HenrikHartiala)

___
*[`migrations/hsl/timetables/1679393480845_add_timetable_version_return_value/up.sql` line 1 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTx2FBIAoNSLIfP1AAA:-NTx2FBIAoNSLIfP1AAB:bdnkx5l) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/1679393480845_add_timetable_version_return_value/up.sql#L1)):*
> ```TSQL
> CREATE SCHEMA return_value;
> ```

Other schemas are created in `migrations/generic/timetables/1000000000000_R_before_migrate/up.sql`, due to how repeatable migrations work (I think). Should this be moved?
___
*[`migrations/hsl/timetables/1679393480845_add_timetable_version_return_value/up.sql` line 15 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwv9go1PSMoxs8j7iv:-NTwv9go1PSMoxs8j7iw:b-62rl7k) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/1679393480845_add_timetable_version_return_value/up.sql#L15)):*
> ```TSQL
> );
> 
> COMMENT ON SCHEMA return_value IS 'This schema is used for all the SLQ functions that need to have a table as return value. Nothing is stored in the tables in this schema.';
> ```

It would also be possible to enforce this with triggers, but probably not worth it.



_Code quote:_
```TSQL
Nothing is stored in the tables in this schema.
```
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 1 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwuMcpEyVYdADDscHu:-NTwuMcpEyVYdADDscHv:b-xx26ty) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L1)):*
> ```TSQL
> CREATE OR REPLACE FUNCTION vehicle_service.get_timetables_and_substitute_operating_days(journey_pattern_uuid uuid, begin_date date, end_date date)
> ```

I would maybe use the actual name of the field here.



_Suggestion:_
```TSQL
journey_pattern_id
```
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 7 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwtCiU1ahLZQIB62qz:-NTwtCiU1ahLZQIB62r-:bfrxgpj) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L7)):*
> ```TSQL
> AS $function$
> SELECT DISTINCT vsf.vehicle_schedule_frame_id, null::uuid, vsf.validity_start, vsf.validity_end, vsf.priority, false, dt.day_type_id 
> 	FROM vehicle_schedule.vehicle_schedule_frame vsf
> ```

Seems to be tabs in these migrations. We typically indent with spaces.
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 12 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwt5Q_EOqxwk2d-L9x:-NTwt5Q_EOqxwk2d-L9y:bejrc56) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L12)):*
> ```TSQL
> 		JOIN vehicle_service.block b USING(vehicle_service_id)
> 		JOIN vehicle_journey.vehicle_journey vj USING(block_id)
> 		JOIN journey_pattern.journey_pattern_ref jpr USING(journey_pattern_ref_id )
> ```

You can use the `vehicle_service.journey_patterns_in_vehicle_service` table here. That way you can drop couple of joins. Something like:
```
SELECT DISTINCT vsf.vehicle_schedule_frame_id, null::uuid, vsf.validity_start, vsf.validity_end, vsf.priority, false, dt.day_type_id
  FROM vehicle_schedule.vehicle_schedule_frame vsf
  JOIN vehicle_service.vehicle_service vs USING (vehicle_schedule_frame_id)
  JOIN service_calendar.day_type dt USING(day_type_id)
  JOIN vehicle_service.journey_patterns_in_vehicle_service jp_in_vs USING(vehicle_service_id)
  WHERE jp_in_vs.journey_pattern_id = journey_pattern_uuid
  AND vsf.validity_end >= begin_date AND vsf.validity_start <= end_date
```
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 18 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwxHhjEt8f4vj_aKB-:-NTwxHhjEt8f4vj_aKB0:b66p2aw) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L18)):*
> ```TSQL
> JOIN  service_calendar.day_type_active_on_day_of_week dtaodow USING(day_type_id)
> GROUP BY day_type_id
> HAVING count(label) = 1)
> ```

The last returned column feels a bit complex/long to be done in SELECT. Might this be neater if done with eg. CTE? Like:
```
WITH 
 active_date_type AS (
   SELECT day_type_id FROM service_calendar.get_active_day_types_for_date(superseded_date) dt
   JOIN  service_calendar.day_type_active_on_day_of_week dtaodow USING(day_type_id)
   GROUP BY day_type_id
   HAVING count(label) = 1
)
SELECT DISTINCT
 null::uuid, sodblt.substitute_operating_day_by_line_type_id, sodblt.superseded_date, sodblt.superseded_date, 23, false,
 active_date_type

```
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 21 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwyTn19LLB7fPIkbL2:-NTwyTn19LLB7fPIkbL3:bz36ix3) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L21)):*
> ```TSQL
> 	FROM journey_pattern.journey_pattern_ref jpr 
> 		JOIN service_calendar.substitute_operating_day_by_line_type sodblt USING(type_of_line)		
> 		WHERE jpr.journey_pattern_id = journey_pattern_uuid AND sodblt.superseded_date >= begin_date AND sodblt.superseded_date <= end_date
> ```

For clarity would be nice to split these to multiple rows:



_Suggestion:_
```TSQL
WHERE jpr.journey_pattern_id = journey_pattern_uuid
AND sodblt.superseded_date >= begin_date
AND sodblt.superseded_date <= end_date
```
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 25 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwye0I9a-YBfsbYkQh:-NTwye0I9a-YBfsbYkQi:b10ibte) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L25)):*
> ```TSQL
> ;
> 
> COMMENT ON FUNCTION vehicle_service.get_timetables_and_substitute_operating_days IS 'Uses return_value.timetable_version as return value. Returns all the corresponding data from 
> ```

Makes the migration file a bit easier to read if you split the first line of comment like this (though we haven't always done that before either).



_Suggestion:_
```TSQL
COMMENT ON FUNCTION vehicle_service.get_timetables_and_substitute_operating_days
IS 'Uses return_value.timetable_version as return value. Returns all the corresponding data from·
```
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 38 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwzQ-A51hWx_QQr4_B:-NTwzQ-A51hWx_QQr4_C:b-n4ulb2) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L38)):*
> ```TSQL
> 	for i IN  1 .. array_upper(journey_pattern_ids, 1)
> 	LOOP 
> 		RETURN query(SELECT * FROM vehicle_service.get_timetables_and_substitute_operating_days(journey_pattern_ids[i], begin_date, end_date));
> ```

There should be a way to do this without a loop.
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 53 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTx-tmlBxQUF0pe0OsG:-NTx-tmlBxQUF0pe0OsH:bmihb3g) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L53)):*
> ```TSQL
> AS $FUNCTION$
> 	DECLARE 
> 	record return_value.timetable_version;
> ```

This isn't used? Also, `record` is a "type" in PostgreSQL, maybe not a good variable name.



_Code quote:_
```TSQL
record
```
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 55 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTx-4r-0PghrfMMv4BD:-NTx-4r-0PghrfMMv4BE:blbun3x) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L55)):*
> ```TSQL
> 	record return_value.timetable_version;
> 	BEGIN
> 		CREATE TEMP TABLE table_holder ON COMMIT DROP AS
> ```

On other temporary tables we have used `ON COMMIT DELETE ROWS`



_Code quote:_
```TSQL
CREATE TEMP TABLE table_holder ON COMMIT DROP AS
```
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 70 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTx0uad2yhZa7Bxfghp:-NTx0uad2yhZa7Bxfghq:b5ymvy2) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L70)):*
> ```TSQL
> 	WHERE priority = 10 AND validity_start <= observation_date AND validity_end >= observation_date AND NOT EXISTS (SELECT * FROM table_holder WHERE day_type_id = subquery.day_type_id AND in_effect = true);
> 		
> 	RETURN query (SELECT * FROM table_holder);
> ```

I would rather try to find another way to implement this function. Running updates on a temporary table and then selecting from it feels like a workaround. Also not parallel safe.

jarkkoka (changes requested):
There are currently no test code to verify that the functions really work. Is it going to be done in a separate PR? If there were some tests, it would be easier to verify that refactoring the SQL code (as suggested by the reviewers) does not brake things.

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-:-NU0Z-937n_WH261KnXK:bcjfg95)* status: 1 of 10 files reviewed, 11 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 0 of 12 files reviewed, 11 unresolved discussions (waiting on @jarkkoka and @Leitsi)

___
*[`migrations/hsl/timetables/1679393480845_add_timetable_version_return_value/up.sql` line 1 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTx2FBIAoNSLIfP1AAA:-NTx7_M360RLzA0uB6uL:b-qx05gq) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/1679393480845_add_timetable_version_return_value/up.sql#L1)):*

Other schemas are created in `migrations/generic/timetables/1000000000000_R_before_migrate/up.sql`, due to how repeatable migrations work (I think). Should this be moved?
</blockquote></details>

Hmm, maybe? Although hmm, should this be in generic? I guess we can use it in both so yeah why not.. check!
___
*[`migrations/hsl/timetables/1679393480845_add_timetable_version_return_value/up.sql` line 15 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwv9go1PSMoxs8j7iv:-NTxAuv03Z_HK9bNP9Jk:bk5ny7e) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/1679393480845_add_timetable_version_return_value/up.sql#L15)):*

It would also be possible to enforce this with triggers, but probably not worth it.
</blockquote></details>

Yeah, Im not that familiar with triggers myself, and we do already have this kind of dummy tables in the database. They just should be moved to this return_value schema.
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 1 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwuMcpEyVYdADDscHu:-NTxBABW1rNOlOvGvSF-:bawymji) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L1)):*

I would maybe use the actual name of the field here.
</blockquote></details>

No, it is for a reason a different name, because this SQL can't use the same name which is used later in the query (`jpr.journey_pattern_id`) Even though there is the prefix `jpr.` it wont work if the parameter name is the same. 
I also checked how this was done earlier and the same method was used (`_id` -> `_uuid`)

It is stupid, but it seems  the only way.

```
CREATE OR REPLACE FUNCTION infrastructure_network.find_point_direction_on_link(point_of_interest public.geography, infrastructure_link_uuid uuid, point_max_distance_in_meters double precision DEFAULT 50.0) RETURNS SETOF infrastructure_network.direction
    LANGUAGE sql STABLE STRICT PARALLEL SAFE
    AS $$
SELECT direction.*
FROM (
    SELECT shape
    FROM infrastructure_network.infrastructure_link
    WHERE infrastructure_link_id = infrastructure_link_uuid
```
This is from `migrations/generic/default/2000000000001_R_after_migrate_create_infrastructure_network/up.sql`
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 7 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwtCiU1ahLZQIB62qz:-NTxBJUj8JRXVroG5b9q:b-80kg7g) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L7)):*

Seems to be tabs in these migrations. We typically indent with spaces.
</blockquote></details>

Ah, must be some file extension specific config etc. don't know why or how these are suddenly not spaces. Luckily this was easy to fix. Should be fixed now
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 12 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwt5Q_EOqxwk2d-L9x:-NTxlJBo5daBLryPwCN7:b-lvytju) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L12)):*

You can use the `vehicle_service.journey_patterns_in_vehicle_service` table here. That way you can drop couple of joins. Something like:
```
SELECT DISTINCT vsf.vehicle_schedule_frame_id, null::uuid, vsf.validity_start, vsf.validity_end, vsf.priority, false, dt.day_type_id
  FROM vehicle_schedule.vehicle_schedule_frame vsf
  JOIN vehicle_service.vehicle_service vs USING (vehicle_schedule_frame_id)
  JOIN service_calendar.day_type dt USING(day_type_id)
  JOIN vehicle_service.journey_patterns_in_vehicle_service jp_in_vs USING(vehicle_service_id)
  WHERE jp_in_vs.journey_pattern_id = journey_pattern_uuid
  AND vsf.validity_end >= begin_date AND vsf.validity_start <= end_date
```
</blockquote></details>

Thanks, I'll use that one
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 18 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwxHhjEt8f4vj_aKB-:-NTxltWS48UrQV2eNk0q:bwi2lw7) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L18)):*

The last returned column feels a bit complex/long to be done in SELECT. Might this be neater if done with eg. CTE? Like:
```
WITH 
 active_date_type AS (
   SELECT day_type_id FROM service_calendar.get_active_day_types_for_date(superseded_date) dt
   JOIN  service_calendar.day_type_active_on_day_of_week dtaodow USING(day_type_id)
   GROUP BY day_type_id
   HAVING count(label) = 1
)
SELECT DISTINCT
 null::uuid, sodblt.substitute_operating_day_by_line_type_id, sodblt.superseded_date, sodblt.superseded_date, 23, false,
 active_date_type

```

</blockquote></details>

Had to google CTE: _Chronic traumatic encephalopathy (CTE) is the term used to describe brain degeneration likely caused by repeated head traumas._ 
:headbang:

But yeah, that could be done with a neater way. Im not sure if I can do it with that plain CTE, because I need to give it the superseded_date as parameter. So should I maybe create a function out of it? 
What do you think?
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 21 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwyTn19LLB7fPIkbL2:-NTxRQJt9jEn2RuvzQhj:b-c0vi5o) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L21)):*

For clarity would be nice to split these to multiple rows:
</blockquote></details>

Will do, thanks! Would be nice to have some linter for these
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 25 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwye0I9a-YBfsbYkQh:-NTxrUHz1exD3G6naYLA:b-dx7h2s) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L25)):*

Makes the migration file a bit easier to read if you split the first line of comment like this (though we haven't always done that before either).
</blockquote></details>

True, Thanks, I'll fix it. Fixed it also in the  already existing before_migrate file.
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 38 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwzQ-A51hWx_QQr4_B:-NTxsNlq8CgTeHYoWMJS:b-xw66g7) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L38)):*

There should be a way to do this without a loop.
</blockquote></details>

This was the first thing that came in to my mind. Is there a downside in using loops?
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 53 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTx-tmlBxQUF0pe0OsG:-NTxtQBh-P52nNFVmWw9:b-wlrz) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L53)):*

This isn't used? Also, `record` is a "type" in PostgreSQL, maybe not a good variable name.
</blockquote></details>

Good one! Must be a leftover from some much older iteration. Removed
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 55 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTx-4r-0PghrfMMv4BD:-NTxtZpGEhwbidPi2inM:b-xk2867) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L55)):*

On other temporary tables we have used `ON COMMIT DELETE ROWS`
</blockquote></details>

Okay nice, this seems better
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 70 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTx0uad2yhZa7Bxfghp:-NTxuxB08XWNUkC8OHnt:bukzc9f) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L70)):*

I would rather try to find another way to implement this function. Running updates on a temporary table and then selecting from it feels like a workaround. Also not parallel safe.
</blockquote></details>

My jacket is a little bit empty about this one, because it is the only way I came up with. Open for any suggestions of course!

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 0 of 13 files reviewed, 11 unresolved discussions (waiting on @jarkkoka and @Leitsi)

___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 38 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTwzQ-A51hWx_QQr4_B:-NU7GBasDtX1z1IjA7vN:b-an5tiu) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L38)):*
<details><summary><i>Previously, HenrikHartiala wrote…</i></summary><blockquote>

This was the first thing that came in to my mind. Is there a downside in using loops?
</blockquote></details>

Thanks for the `unnest` suggestion! I think I got it working. I used this `unnest` and `CROSS JOIN LATERAL`. Does this look correct now?

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 0 of 13 files reviewed, 11 unresolved discussions (waiting on @jarkkoka and @Leitsi)

___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 55 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NTx-4r-0PghrfMMv4BD:-NUVTcrI3eYMu-0rwwju:b-ebpxoa) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/329c4764a81cddea3b6b6b527e1ad45b71869e8b/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L55)):*
<details><summary><i>Previously, HenrikHartiala wrote…</i></summary><blockquote>

Okay nice, this seems better
</blockquote></details>

Hmm, after a while I realised this does not work for some reason. After first fetch the result of this function is not correct anymore. I think I need to revert this back to `ON COMMIT DROP`.

jarkkoka (commented):
> Reviewed 4 of 10 files at r1, 3 of 5 files at r2, 1 of 2 files at r3, 3 of 4 files at r5, 1 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 12 of 13 files reviewed, 13 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`-- commits` line 9 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NUvHwo91-kpsrK0uh4A):*
Btw, do there already exist backlog tickets about the future TBD?
___
*[`migrations/generic/timetables/1000000000000_R_before_migrate/up.sql` line 40 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVn97g28fSb4sExQF1t:-NVn97g28fSb4sExQF1u:babb2iy) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/generic/timetables/1000000000000_R_before_migrate/up.sql#L40)):*
> ```TSQL
> CREATE SCHEMA IF NOT EXISTS return_value;
> COMMENT ON SCHEMA return_value 
> IS 'This schema is used for all the SLQ functions that need to have a table as return value. Nothing is stored in the tables in this schema.';
> ```

typo: `SLQ` -> `SQL`

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 12 of 13 files reviewed, 14 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 16 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVn9g9v-UcukAPyX02i:-NVn9g9v-UcukAPyX02j:b-yhymvu) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L16)):*
> ```TSQL
>   AND vsf.validity_start <= end_date
>   AND vsf.priority != 40 -- ignore staging priorities
> union 
> ```

According to project convention all SQL keywords are in uppercase except for column types: `union` -> `UNION`

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 12 of 13 files reviewed, 15 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 15 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnAQxe2o3nohwHhRVu:-NVnAQxe2o3nohwHhRVv:bu1vo1d) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L15)):*
> ```TSQL
>   AND vsf.validity_end >= start_date 
>   AND vsf.validity_start <= end_date
>   AND vsf.priority != 40 -- ignore staging priorities
> ```

There is already a function `internal_utils.const_priority_draft` which concerns the networks & route model.

We have so far avoided to hard-code these magic values into functions. I think this value could be returned by an `internal_utils` function. Maybe need to resolve naming conflict at the same time (networkdb vs timetablesdb).

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 12 of 13 files reviewed, 16 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 17 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnBnyE2XI2rSLFsx8o:-NVnBnyE2XI2rSLFsx8p:b-42llaf) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L17)):*
> ```TSQL
>   AND vsf.priority != 40 -- ignore staging priorities
> union 
> SELECT DISTINCT null::uuid, sodblt.substitute_operating_day_by_line_type_id, sodblt.superseded_date, sodblt.superseded_date, 23, false, (
> ```

An uncommented magic value hard-coded into the query. Would prefer to see a function call here that returns this.

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 12 of 13 files reviewed, 17 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 6 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnCvyt2reqsTcw6se6:-NVnCvyt2reqsTcw6se7:b-i8r5e) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L6)):*
> ```TSQL
>  STABLE
> AS $function$
> SELECT DISTINCT vsf.vehicle_schedule_frame_id, null::uuid, vsf.validity_start, vsf.validity_end, vsf.priority, false, dt.day_type_id 
> ```

put nulls in uppercase (`null` -> `NULL`). Multiple occurrences in this file.

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 12 of 13 files reviewed, 18 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 18 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnGAVJ2tT5SyvTEcxU:-NVnGAVJ2tT5SyvTEcxV:b1kmpg1) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L18)):*
> ```TSQL
> union 
> SELECT DISTINCT null::uuid, sodblt.substitute_operating_day_by_line_type_id, sodblt.superseded_date, sodblt.superseded_date, 23, false, (
>   SELECT day_type_id FROM service_calendar.get_active_day_types_for_date(sodblt.superseded_date) dt
> ```

It would help readability to start "FROM" on a new line.

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 12 of 13 files reviewed, 17 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 36 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnOO1o753vIxuAqofT:-NVnOO1o753vIxuAqofU:b-bfbgfm) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L36)):*
> ```TSQL
> but leaves out staging priorities completely.
> This function will also match the day type for the substitute day by using the superseded date.
> In the return values, either the vehicle_schedule_frame_id OR substitute_operating_day_by_line_type_id will be NULL.';
> ```

I think it needs to be commented why this function always returns `in_effect = false` since it is not obvious.

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 12 of 13 files reviewed, 18 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`migrations/hsl/timetables/1679393480845_add_timetable_version_return_value/up.sql` line 8 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnOvDT6Xr46L9Diokg:-NVnOvDT6Xr46L9Diokh:bdr9jlf) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/1679393480845_add_timetable_version_return_value/up.sql#L8)):*
> ```TSQL
>   validity_end date NOT NULL,
>   priority integer NOT NULL,
>   in_effect boolean NOT NULL,
> ```

Please document what `in_effect` means.

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 12 of 13 files reviewed, 19 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 43 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnPGrp--II4DqTcAim:-NVnPGrp--II4DqTcAin:bfsdrac) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L43)):*
> ```TSQL
>  STABLE
> AS $FUNCTION$
>   WITH
> ```

No indent needed before `WITH`.

jarkkoka (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 12 of 13 files reviewed, 19 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 12 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnGW5K0o4XNlU01FXm:-NVnPbswAwHy3vIVFJHQ:bozfiyo) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L12)):*

I think `ANY` would be clearer (think it works here). Same thing a bit more below.
</blockquote></details>

Yes, agree. That actually works.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 9 of 13 files reviewed, 19 unresolved discussions (waiting on @jarkkoka and @Leitsi)

___
*[`-- commits` line 9 at r5](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NUvHwo91-kpsrK0uh4A):*

Btw, do there already exist backlog tickets about the future TBD?
</blockquote></details>

Thanks. I just createad a ticket about this. #1287
___
*[`migrations/generic/timetables/1000000000000_R_before_migrate/up.sql` line 40 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVn97g28fSb4sExQF1t:-NVnUPHBAPcad_o93BAH:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/generic/timetables/1000000000000_R_before_migrate/up.sql#L40)):*

typo: `SLQ` -> `SQL`
</blockquote></details>

:+1: 
___
*[`migrations/hsl/timetables/1679393480845_add_timetable_version_return_value/up.sql` line 8 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnOvDT6Xr46L9Diokg:-NVnXByC5GtPdVd77gxv:balam34) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/1679393480845_add_timetable_version_return_value/up.sql#L8)):*

Please document what `in_effect` means.
</blockquote></details>

Added a description with examples to the table comment.
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 5 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnCg5s5MyTzwHeEMt-:-NVnYC9r8ESL4faZSW9W:bkjxpv0) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L5)):*

You could just write these as `AS $$`, since there's no need to use them with a named tag here. https://stackoverflow.com/a/12172353
</blockquote></details>

alright
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 6 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnCvyt2reqsTcw6se6:-NVnYy301jBdU1rzvi_-:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L6)):*

put nulls in uppercase (`null` -> `NULL`). Multiple occurrences in this file.
</blockquote></details>

:+1: 
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 10 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnBs6g14sxuamYOzaa:-NVnZHd65ejG9cO-XhMh:b-hpt5t3) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L10)):*

block isn't actually used, can omit?
</blockquote></details>

true, not actually sure if it was needed before I used the journey_patterns_in_vehicle_service's even. Good spot!
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 12 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnGW5K0o4XNlU01FXm:-NVnZfhg5zfjeWCIYtVN:b-l0sg85) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L12)):*

Yes, agree. That actually works.
</blockquote></details>

ah, nice! Thats much better.
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 13 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnD1rL-Kh5v5sgrexH:-NVnlHen2YlXN_PZsKks:bmgsh3v) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L13)):*

Quite a few trailing spaces in this file, could clean those up.
</blockquote></details>

Ah right, coz there is no on-save formatting and did not have any vscode default for this. Added `"files.trimTrailingWhitespace": true` to my settings file and now this should be fixed.
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 15 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnAQxe2o3nohwHhRVu:-NVnlXkM6iyVtxkyTEpK:bt36i4y) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L15)):*

There is already a function `internal_utils.const_priority_draft` which concerns the networks & route model.

We have so far avoided to hard-code these magic values into functions. I think this value could be returned by an `internal_utils` function. Maybe need to resolve naming conflict at the same time (networkdb vs timetablesdb).
</blockquote></details>

We have used "timetablesPriority" in the UI, so maybe here as well `const_timetables_priority_draft` and so on.
But yeah, did not know about these, thanks! Copypasted the functions otherwise, but used the const_timetables_ prefix.
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 16 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVn9g9v-UcukAPyX02i:-NVnnNzI4H1bjNAm2hsX:b6lkke1) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L16)):*

According to project convention all SQL keywords are in uppercase except for column types: `union` -> `UNION`
</blockquote></details>

Yeah, there is no formatter available here and for some reason I could not get any installed either. Dont know what is the problem but for some reasun sqlfluff etc, just did not work at least yet. So I'll try to fix these by hand then. 
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 17 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnBnyE2XI2rSLFsx8o:-NVntyzuAuAOng5LS9mm:b3lly8n) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L17)):*

It would really help readability if all the values in the return column were each on their own row and if the constant literals were followed by an explanatory SQL comment (like what that `false` means). Now, I have to constantly switch to separate file to see what the columns of return value table are.
</blockquote></details>

Great, added comments and also function for the magic number 23
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 18 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnGAVJ2tT5SyvTEcxU:-NVnuKdp3v3ldBNXjup8:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L18)):*

It would help readability to start "FROM" on a new line.
</blockquote></details>

:+1: 
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 36 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnOO1o753vIxuAqofT:-NVnuwjIARgtQuis7pt7:b-d8r4ej) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L36)):*

I think it needs to be commented why this function always returns `in_effect = false` since it is not obvious.
</blockquote></details>

Check, added comments 
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 43 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnPGrp--II4DqTcAim:-NVnv3DFCT2iAxQ2cFDS:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L43)):*

No indent needed before `WITH`.
</blockquote></details>

:+1: 
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 47 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnDrY71qkH2x3TqttO:-NVnvFQk8FzgOPcHwHLF:bb6mt9b) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L47)):*

Confusingly placed closing bracket. I'd either move this to end of previous line or split the parameters on separate line, like:
```
  SELECT gen_random_uuid() AS id, * FROM
  vehicle_service.get_timetables_and_substitute_operating_days(
    journey_pattern_ids, start_date, end_date
  )
```

</blockquote></details>

:+1:
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 50 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnEEefEPvACd8ddMQj:-NVnvYDS9FURthofg0S7:b-8fx2u7) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L50)):*

Couple levels of extra indentation here.
</blockquote></details>

:+1: 
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 57 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnFqMICQmuu28_MBG7:-NVnvbRH6S2EoysjnZyH:bxgz6jt) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L57)):*

Maybe add those `const_priority_draft` calls here as well.
</blockquote></details>

yes, added everywhere were they were used.
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 65 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnFcI9C-0j1JUZ2u1Z:-NVnvvkQ1vqr-uTz_PTy:bg0him0) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L65)):*

Overriding with staging is still fine though? Just checking.

EDIT: Ah, this is handled by the `get_timetables_and_substitute_operating_days` not returning staging priorities. Maybe slightly confusing having the priority handling split between these two functions.
</blockquote></details>

Yeah, I'm not sure which would be the best way here. I think I just didn't consider the "staging" as priority at all, and that's why it is already left out in the other function. Here we have the real priorities left (although don't consider the draft as a priority either :D, oh well..). But we do want to show the drafts, even though they are separate concept from the other priorities. In my opinion this is exactly the reason why neither "draft" nor "staging" should not even be priorities, since they are not priorities, they are separate concepts.

TL;DR: 
* We want to ignore staging priorities completely, because they are not a thing.
* We want to show the draft priorities, but they can't be part of the in_effect calculation.

I'll more descriptive comments about this. 
Also will take suggestions.
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 71 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnFCI5CGhQlF2FCj_j:-NVo-QZ5AsOrdVcoIdTA:b-i919ue) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L71)):*

Brackets placement a bit confusing: closing and opening at different indentation levels. I would indent like this: 
</blockquote></details>

Tried to fix these now.

jarkkoka (commented):
> Reviewed 3 of 3 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: all files reviewed, 11 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`migrations/hsl/timetables/1679393480845_add_timetable_version_return_value/up.sql` line 8 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnOvDT6Xr46L9Diokg:-NW1ebON5s8Wya9kR3Oh:b59xw24) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/1679393480845_add_timetable_version_return_value/up.sql#L8)):*
<details><summary><i>Previously, HenrikHartiala wrote…</i></summary><blockquote>

Added a description with examples to the table comment.
</blockquote></details>

Ok, thanks. Good coverage! :+1: 
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 43 at r8](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NW2cErs6ntffzBaeDwy:-NW2cErs6ntffzBaeDwz:b-ti0qv) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/a4e5d737afd654d669b0e9dc11e6141a3cf95376/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L43)):*
> ```TSQL
>   sodblt.superseded_date, -- validity start is the superseded date for substitute operating day by line type
>   sodblt.superseded_date, -- validity end is the superseded date for substitute operating day by line type
>   internal_utils.const_timetables_priority_substitute_by_line_type(), --priority for these are substitute by line type
> ```

would need a space after double dash required for a valid SQL comment (`-- `)

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: all files reviewed, 10 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 13 at r6](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NVnD1rL-Kh5v5sgrexH:-NW2fah_EtEFBXKysInj:b2nhoia) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/43525c900dcb66b3792d03f984fb4ee848bde208/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L13)):*
<details><summary><i>Previously, HenrikHartiala wrote…</i></summary><blockquote>

Ah right, coz there is no on-save formatting and did not have any vscode default for this. Added `"files.trimTrailingWhitespace": true` to my settings file and now this should be fixed.
</blockquote></details>

TIL

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: all files reviewed, 11 unresolved discussions (waiting on @HenrikHartiala and @Leitsi)

___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 37 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NW2g3UH4u4chzTWsoMM:-NW2g3UH4u4chzTWsoMN:b-9c9902) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/858eb9fef3bbaf2494494ed4370d3284579e04eb/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L37)):*
> ```TSQL
>     AND vsf.validity_start <= end_date
>     AND vsf.priority != internal_utils.const_timetables_priority_staging() -- ignore staging priorities completely, since they are not shown anywhere
> union
> ```

should be uppercase `UNION`

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 12 of 13 files reviewed, 11 unresolved discussions (waiting on @jarkkoka and @Leitsi)

___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 43 at r8](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NW2cErs6ntffzBaeDwy:-NW6hLJQ8Jmok87HfQrO:be6vibm) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/a4e5d737afd654d669b0e9dc11e6141a3cf95376/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L43)):*

would need a space after double dash required for a valid SQL comment (`-- `)
</blockquote></details>

Ah, thanks. space should be added to everywhere now
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 37 at r9](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NW2g3UH4u4chzTWsoMM:-NW6gai4AH-5-VJ0rND5:bkwsj7b) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/858eb9fef3bbaf2494494ed4370d3284579e04eb/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L37)):*

should be uppercase `UNION`
</blockquote></details>

Thanks, next time I'm just going to use the dbeaver to format my SQL :D

Leitsi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: 12 of 13 files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala and @jarkkoka)

___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 13 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NW6uA8J3WkyVZ3adCFQ:-NW6uA8J3WkyVZ3adCFR:bmi7h0d) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/e847cb48d40ea0b4841cf2aeadd2b540a6c64b90/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L13)):*
> ```TSQL
>   AS $$SELECT 30$$;
> 
> CREATE OR REPLACE FUNCTION internal_utils.const_timetables_priority_staging() RETURNS integer
> ```

Maybe move these immutable functions to `migrations/generic/timetables/1632000000000_create_immutable_internal_utils`. We might want to use them elsewhere as well.

jarkkoka (commented):
> Reviewed 3 of 3 files at r11, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: all files reviewed, 3 unresolved discussions (waiting on @HenrikHartiala)

___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 80 at r12](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NW7mPCgCgxEZIlzAWSP:-NW7mPCgCgxEZIlzAWSQ:b-pbj9u0) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/dfb7b987d47cf2729b59f3ddd177dbfce4e068ac/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L80)):*
> ```TSQL
>         FROM timetable_versions overriding_tv
>         WHERE
>           overriding_tv.id != timetable_versions.id
> ```

This query appears to be `O(n^2)` in complexity, so it may not perform well on large datasets. However, I couldn't easily come up with a more efficient version without overcomplicating things. Hopefully this won't used be to handle so many journey patterns at once.

Leitsi (changes requested):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: all files reviewed, 4 unresolved discussions (waiting on @HenrikHartiala)

___
*[`migrations/generic/timetables/1632000000000_create_immutable_internal_utils/up.sql` line 13 at r12](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NW7pTgIEsyOYFa4TZkM:-NW7pTgIEsyOYFa4TZkN:bj7h9k5) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/dfb7b987d47cf2729b59f3ddd177dbfce4e068ac/migrations/generic/timetables/1632000000000_create_immutable_internal_utils/up.sql#L13)):*
> ```TSQL
>   AS $$SELECT 25$$;
> 
> CREATE OR REPLACE FUNCTION internal_utils.const_timetables_priority_draft() RETURNS integer
> ```

Isn't this exactly the same as `internal_utils.const_priority_draft()` above? These seem to refer to same priority fields as well. Maybe just remove this one and rename others similarly to the old function, `const_priority_xxx`.

HenrikHartiala (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178)* status: all files reviewed, 3 unresolved discussions (waiting on @jarkkoka and @Leitsi)

___
*[`migrations/generic/timetables/1632000000000_create_immutable_internal_utils/up.sql` line 13 at r12](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NW7pTgIEsyOYFa4TZkM:-NWBaowwEVwwtc4tkdCG:bvtn0kh) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/dfb7b987d47cf2729b59f3ddd177dbfce4e068ac/migrations/generic/timetables/1632000000000_create_immutable_internal_utils/up.sql#L13)):*

Isn't this exactly the same as `internal_utils.const_priority_draft()` above? These seem to refer to same priority fields as well. Maybe just remove this one and rename others similarly to the old function, `const_priority_xxx`. 
</blockquote></details>

It is the same, but the value could be different. In the end `Timetables Priority` and `Priority` are two different things and we also at some point discussed that the values could also differ with one hundred, e.g. `standard priority = 10`, `timetables standard priority 110` and so on.
So I would want to separate them with the `timetables_` prefix, and also to be safe, I would leave the duplicates there, in case we decide to change the 
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 13 at r10](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NW6uA8J3WkyVZ3adCFQ:-NWBdG2l9GyUvuOt9YRv:b-896fix) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/e847cb48d40ea0b4841cf2aeadd2b540a6c64b90/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L13)):*

Maybe move these immutable functions to `migrations/generic/timetables/1632000000000_create_immutable_internal_utils`. We might want to use them elsewhere as well.
</blockquote></details>

Done.
___
*[`migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql` line 80 at r12](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-NW7mPCgCgxEZIlzAWSP:-NWBdRBX03datWuCBvaa:b4eny11) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/dfb7b987d47cf2729b59f3ddd177dbfce4e068ac/migrations/hsl/timetables/2000000000103_R_after_migrate_create_get_timetable_versions/up.sql#L80)):*

This query appears to be `O(n^2)` in complexity, so it may not perform well on large datasets. However, I couldn't easily come up with a more efficient version without overcomplicating things. Hopefully this won't used be to handle so many journey patterns at once.
</blockquote></details>

:+1: We'll come back to it if it becomes a problem

Leitsi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/178#-:-NWCPW8l0g8675CitXL1:bnfp4nl)* status: 12 of 16 files reviewed, 1 unresolved discussion (waiting on @jarkkoka)

PR #182, opened 2023-05-09
Add script for generating dumps on each commit with git rebase

Leitsi:
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/182)

### DIFF
--- a/scripts/amend-generate-dumps.sh
+++ b/scripts/amend-generate-dumps.sh
@@ -1,0 +1,75 @@
+#!/bin/sh
+
+# Recreates dependencies, generates database dumps and adds them to current commit.
+#
+# NB: because the dependencies are recreated, the testdb gets erased.
+#
+# Intended to be used with git rebase, eg:
+#   git rebase origin/main --exec=./scripts/amend-generate-dumps.sh
+#
+# Can also be used outside rebase, in which case the changes get amended to current commit.
+#
+# Known caveats:
+# - Does not detect if migrations ran successfully.
+#   This might result in incorrect dumps being created (most SQL shown as removed).
+#   Thus the commits should still be manually checked in case there are any clear anomalies.
+
+set -eu
+
+# allow running from any working directory
+WD=$(dirname "$0")
+cd "${WD}"
+
+if ! git diff --quiet; then
+  echo "There are uncommitted changes. Please commit or stash them before running the script."
+  exit 1
+fi
+
+# Make sure the generic schema is selected, otherwise dumps are not generated correctly. Should be by default.
+SELECTED_MIGRATION_DIRECTORY=$(grep "migrations_directory: " ../config.yaml | cut -d ":" -f2 | tr -d " ")
+if [ "$SELECTED_MIGRATION_DIRECTORY" != "migrations/generic" ]; then
+  echo "Generic schema not selected. Aborting."
+  exit 1
+fi
+
+# Generate dumps expects generic image to have been built.
+echo "Setting docker target."
+./set-docker-target.sh generic
+
+echo "Restarting dependencies to recreate test database."
+bash ./stop-dependencies.sh
+bash ./start-dependencies.sh --detach
+
+# It is important that DB is all ready and migrations have been ran, before we try to generate the dumps.
+echo "Waiting for docker images to be ready."
+until [ "`docker inspect -f {{.State.Health.Status}} testdb`"=="healthy" ]; do
+  sleep 0.1;
+done;
+until [ "`docker inspect -f {{.State.Health.Status}} hasura`"=="healthy" ]; do
+  sleep 0.1;
+done;
+# Merely being healthy doesn't seem to be enough: running migrations might not have finished.
+# TODO: is there a better way?
+sleep 5;
+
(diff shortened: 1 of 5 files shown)


Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/182)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @culka)

___
*[`scripts/amend-generate-dumps.sh` line 12 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/182#-NV3xA4BBsEC28Yy8yeE:-NV3yMK6-zMfB5WK91hq:b6wvfev) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/45799ebb6ff0873333a5728a280c8f19bffbbad8/scripts/amend-generate-dumps.sh#L12)):*

Should this (or maybe `generate-dumps.sh` rather) warn if you're not in generic schema and have HSL specific migrations?

Not a blocker but just pondering.
</blockquote></details>

Possibly. Should maybe exit with error in that case.
___
*[`scripts/amend-generate-dumps.sh` line 36 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/182#-NV3vEfF8MQ92Ty7rFNE:-NV3ymw1-WTHxUEB6o5U:b4c9lyr) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/45799ebb6ff0873333a5728a280c8f19bffbbad8/scripts/amend-generate-dumps.sh#L36)):*

ahve
</blockquote></details>

:thumbsup: 
___
*[`scripts/amend-generate-dumps.sh` line 45 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/182#-NV3toXi4dty9oEXWQkd:-NV3zQ544SrrVhp5tBxJ:bjrjn2f) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/45799ebb6ff0873333a5728a280c8f19bffbbad8/scripts/amend-generate-dumps.sh#L45)):*

You can check something like what is in `jore4-ui` startup scripts

`  if [[ $(docker exec $1 psql $DB_CONNECTION_STRING -AXqtc "SELECT EXISTS (SELECT FROM pg_tables WHERE schemaname = 'service_pattern' AND tablename = 'distance_between_stops_calculation');") = "t" ]]; then`
Would check if the latest generic migration that adds a table has been executed, but this requires upkeep when we add new migrations. 

What could be used for this is a repeatable migration that sets a value once it is executed and which is cleared in before migrate?
</blockquote></details>

Hmm, maybe. The repeatable migration solution sounds fine and not too much of an overkill for this, even though would be nicer to fix this outside DB.
___
*[`scripts/set-docker-target.sh` line 8 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/182#-NV3w53qDInTzihfIm-Y:-NV3yyg74Vmg_YjeV1GY:b387xvy) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/45799ebb6ff0873333a5728a280c8f19bffbbad8/scripts/set-docker-target.sh#L8)):*

`docket`

If you want to avoid the mistake I made in the script below, use `basename $0` for the name.
</blockquote></details>

:facepalm: :thumbsup: 
___
*[`scripts/use-specific-schema.sh` line 8 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/182#-NV3UgOz8ZWHtRutT0iF:-NV3z33O280aiaZ2uTsg:b2eidal) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/45799ebb6ff0873333a5728a280c8f19bffbbad8/scripts/use-specific-schema.sh#L8)):*

Whoops. I should have used `basename $0` here...
</blockquote></details>

:ok_hand:  will change.

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/182)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @culka)

___
*[`scripts/amend-generate-dumps.sh` line 45 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/182#-NV3toXi4dty9oEXWQkd:-NVAdHPz0T-e-72eQWxR:bwbok6b) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/45799ebb6ff0873333a5728a280c8f19bffbbad8/scripts/amend-generate-dumps.sh#L45)):*

Hmm, maybe. The repeatable migration solution sounds fine and not too much of an overkill for this, even though would be nicer to fix this outside DB.
</blockquote></details>

Spent a while investigating this but couldn't get a working solution. SET commands only work for same session, which is not sufficient. Could always implement with some custom table + inserting a value but feels overkill.

Also tried digging at Hasura tables, but there doesn't seem to be any clear data about ran migrations stored in any of those.

So, going with this neat sleep hack "for now"

culka (approved):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/182)* status: 0 of 5 files reviewed, all discussions resolved

PR #190, opened 2023-07-10
Remove timetables-namespace

Perholehto:
Adding namespace for Timetables breaks transaction when used from Apollo Client...

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/190)


Perholehto:
There is still discussion about how this should be done or should it even be done.

PR #212, opened 2023-10-31
Validate sequential integrity within vehicle_services: don't allow overlapping blocks of journeys

Leitsi:
Resolves https://github.com/HSLdevcom/jore4/issues/1483

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212)


culka (commented):
> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212)* status: all files reviewed, 1 unresolved discussion (waiting on @Leitsi)

___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 137 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-Ni9ifnSBycjbHIdA5CW:-Ni9ifnSBycjbHIdA5CX:blp3amn) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/209a328c1f9baa15f7289996e428653c6bed8f01/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L137)):*
> ```TSQL
>     ),
>     with_inconsistencies AS (
>       SELECT
> ```

I would like if there was a comment here to explain how these are constructed and how the comparison is done. It took a while to parse what happened and if someone needs to do the same again a comment would help.

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212)* status: 8 of 19 files reviewed, 1 unresolved discussion (waiting on @culka)

___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 137 at r1](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-Ni9ifnSBycjbHIdA5CW:-NiDk4dQ4NZpV79wYNhy:b-4todcs) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/209a328c1f9baa15f7289996e428653c6bed8f01/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L137)):*

I would like if there was a comment here to explain how these are constructed and how the comparison is done. It took a while to parse what happened and if someone needs to do the same again a comment would help.
</blockquote></details>

:ok: tried to clarify this a bit: reordered and reformatted the selected rows here, added comments, tried to clarify naming.

jarkkoka (changes requested):
> Reviewed 12 of 12 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212)* status: all files reviewed, 14 unresolved discussions (waiting on @culka and @Leitsi)

___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 37 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-NiiIbHw8sDoxLP5z8dv:-NiiIbHw8sDoxLP5z8dw:b-uc54jg) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L37)):*
> ```TSQL
>       vehicle_service_id,
>       MIN(tpt.departure_time) OVER (PARTITION BY vehicle_journey_id) AS journey_first_stop_departure,
>       MAX(tpt.arrival_time) OVER (PARTITION BY vehicle_journey_id) AS journey_last_stop_arrival,
> ```

I think you should coalesce `arrival_time` with `passing_time` since `arrival_time` is nullable. Now, you could get the arrival time of a stop point that is not the last in the journey.

The same may apply to departure time above. 
___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 57 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-NiiU4fu1J2l-MxQK1te:-NiiU4fu1J2l-MxQK1tf:b-kf1w1s) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L57)):*
> ```TSQL
>       *,
>       journey_first_stop_departure - layover_time AS journey_start,
>       journey_last_stop_arrival + turnaround_time AS journey_end
> ```

What/where is the specification for determining journey start and end time?

To me it would be more intuitive to use `turnaround_time` in determining the journey start time. In Hastus import files, the field is called `trp_lay_min_strt`. In Finnish, it is "Terminaaliaika".

Respectively, t would be more intuitive to me to use `layover_time` in determining the journey end time. In Hastus import files, the field is called `trp_lay_min_end`. In Finnish, it is "Elpymisaika".

Are you sure about the code is correct here?
___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 112 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-NiiyfSSDugnP7_sZVNX:-NiiyfSSDugnP7_sZVNY:b-2xbedx) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L112)):*
> ```TSQL
>     -- Eg. if block_id is present in modified_block, it's parent vehicle_service will be selected.
>     WITH modified_vehicle_service_ids AS (
>       SELECT vehicle_service_id
> ```

It seems to that `SELECT DISTINCT` could be used here to return unique set of identifiers.
___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 155 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-NiizhlE8aLpPc--31ym:-NiizhlE8aLpPc--31yn:b-viedq5) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L155)):*
> ```TSQL
>         (
>           in_temporal_order.block_id <> next_time.block_id
>           AND in_temporal_order.block_end > next_time.block_start
> ```

Because of the `>` operator, a block can end in the same minute as the next one starts. I think this is allowed?
___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 159 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-Nij-2OzC41A7dn6dBAq:-Nij-2OzC41A7dn6dBAr:b-ru4dsa) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L159)):*
> ```TSQL
>         (
>           in_temporal_order.vehicle_journey_id <> next_time.vehicle_journey_id
>           AND in_temporal_order.journey_end > next_time.journey_start
> ```

Because of the `>` operator, a journey can end in the same minute as the next one starts. I think this is allowed?
___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 164 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-Nij-UXs1QrCsU49mDNx:-Nij-UXs1QrCsU49mDNy:bhphhtm) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L164)):*
> ```TSQL
>       JOIN in_temporal_order next_time
>         ON in_temporal_order.vehicle_service_id = next_time.vehicle_service_id
>         AND next_time.order_number = (in_temporal_order.order_number + 1)
> ```

The last item in the sorted `in_temporal_order` set is discarded according to the join condition. I think the logic works well with it.
___
*[`test/hasura/generic/timetablesdb/datasets/types.ts` line 102 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-Nij1gdg1TMVZ5QXtQaT:-Nij1gdg1TMVZ5QXtQaU:b-3nfd6p) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/test/hasura/generic/timetablesdb/datasets/types.ts#L102)):*
> ```XML
>   journey_pattern_ref_id: UUID;
>   layover_time: Duration | null;
>   turnaround_time: Duration | null;
> ```

I would order `turnaround_time` before `layover_time` for clarity.
___
*[`test/hasura/generic/timetablesdb/datasets/types.ts` line 110 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-Nij1nWj5XrGwH2hhdaW:-Nij1nWj5XrGwH2hhdaX:b-3nfd6p) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/test/hasura/generic/timetablesdb/datasets/types.ts#L110)):*
> ```XML
>   'journey_pattern_ref_id',
>   'layover_time',
>   'turnaround_time',
> ```

I would order `turnaround_time` before `layover_time` for clarity.
___
*[`test/hasura/generic/timetablesdb/datasets/defaultSetup/vehicle-journeys.ts` line 11 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-Nij1pVx3Z6HWaTy14de:-Nij1pVx3Z6HWaTy14df:b-xa78rr) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/test/hasura/generic/timetablesdb/datasets/defaultSetup/vehicle-journeys.ts#L11)):*
> ```XML
>     journey_pattern_ref_id:
>       journeyPatternRefsByName.route123Outbound.journey_pattern_ref_id,
>     layover_time: null,
> ```

I would change the order of these two properties for all occurrences in the file.
___
*[`test/hasura/generic/timetablesdb/integration-tests/data-integrity/vehicle-service/sequential-integrity.spec.ts` line 81 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-Nij2Qpb5z9-tluIBPHY:-Nij2Qpb5z9-tluIBPHZ:blxut6k) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/test/hasura/generic/timetablesdb/integration-tests/data-integrity/vehicle-service/sequential-integrity.spec.ts#L81)):*
> ```XML
>     direction?: RouteDirection;
>     layover?: string | null;
>     turnaround?: string | null;
> ```

I would change the order of these two properties for the sake of logical consistency and clarity (for all occurrences in this file).
___
*[`test/hasura/hsl/timetablesdb/datasets/defaultSetup/vehicle-journeys.ts` line 16 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-Nij24-l2qNqmw-_lVUi:-Nij24-l2qNqmw-_lVUj:bf2c3ay) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/test/hasura/hsl/timetablesdb/datasets/defaultSetup/vehicle-journeys.ts#L16)):*
> ```XML
>       journeyPatternRefsByName.route123Inbound.journey_pattern_ref_id,
>     layover_time: null,
>     turnaround_time: null,
> ```

I would change the order of these two new properties for all occurences in this file.
___
*[`test/hasura/timetables-data-inserter/generic/models/vehicle-journey.ts` line 63 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-Nij2CwJ0Omu5bdVvQ3A:-Nij2CwJ0Omu5bdVvQ3B:b-w28r4l) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/test/hasura/timetables-data-inserter/generic/models/vehicle-journey.ts#L63)):*
> ```XML
>   return {
>     layover_time: null,
>     turnaround_time: null,
> ```

I would change the order of these two new properties.
___
*[`test/hasura/generic/timetablesdb/integration-tests/data-integrity/vehicle-service/journey-patterns-in-vehicle-service.spec.ts` line 86 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-Nij1xpY9pp27NuHremu:-Nij1xpY9pp27NuHremv:b-oij1dv) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/test/hasura/generic/timetablesdb/integration-tests/data-integrity/vehicle-service/journey-patterns-in-vehicle-service.spec.ts#L86)):*
> ```XML
>           journeyPatternRefsByName.route123Outbound.journey_pattern_ref_id,
>         layover_time: null,
>         turnaround_time: null,
> ```

I would change the order of the two new properties for both two occurrences.

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212)* status: all files reviewed, 14 unresolved discussions (waiting on @culka and @jarkkoka)

___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 155 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-NiizhlE8aLpPc--31ym:-Nino8lb4yiRxMURkNtB:bev60w1) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L155)):*

Because of the `>` operator, a block can end in the same minute as the next one starts. I think this is allowed?
</blockquote></details>

It is now!
But to be serious, good question. 

I would say typically time ranges work like this, that is, start closed end open (eg. PostgreSQL handles interval comparison like this by default). So to me this current version would make sense.

**But**, another similar case we already have is vehicle schedule frame validity ranges. Those have been handled as fully closed.
Relevant commit: [c3344921f7156c68afe2885117fe252e91565195](https://github.com/HSLdevcom/jore4-hasura/commit/c3344921f7156c68afe2885117fe252e91565195)

I suppose it's also relevant how Hastus handles these.

In practice this probably makes no difference: typically timetables are designed so that there's some breathing room anyways, so I wonder if we would ever even encounter these cases.
___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 164 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-Nij-UXs1QrCsU49mDNx:-NintrPk9spog1g1caii:b4wr4yu) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L164)):*

The last item in the sorted `in_temporal_order` set is discarded according to the join condition. I think the logic works well with it.
</blockquote></details>

Yep. The main point here is checking the integrity, that is, that each row (block or journey) does not overlap with the next one. For the last one there is no next one to compare against, there's no next row whose start it could overlap with. Previous row already checks overlap issues with the start of this last row that is discarded. So all is good.

Of course we could change things around a bit, do a `LEFT JOIN` instead, but that would just result in `block_overlaps_with_next` and `journey_overlaps_with_next` evaluating to `false` anyway, so no reason to complicate the logic because of that.

But maybe I'll add a comment that this has been thought of.

jarkkoka (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212)* status: all files reviewed, 14 unresolved discussions (waiting on @culka and @Leitsi)

___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 155 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-NiizhlE8aLpPc--31ym:-Nio207R2DrZJO6SiNmx:b-ohouj3) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L155)):*

It is now!
But to be serious, good question. 

I would say typically time ranges work like this, that is, start closed end open (eg. PostgreSQL handles interval comparison like this by default). So to me this current version would make sense.

**But**, another similar case we already have is vehicle schedule frame validity ranges. Those have been handled as fully closed.
Relevant commit: [c3344921f7156c68afe2885117fe252e91565195](https://github.com/HSLdevcom/jore4-hasura/commit/c3344921f7156c68afe2885117fe252e91565195)

I suppose it's also relevant how Hastus handles these.

In practice this probably makes no difference: typically timetables are designed so that there's some breathing room anyways, so I wonder if we would ever even encounter these cases.
</blockquote></details>

This is a bit complicated when the turnaround time and layover time are zeros. Vehicle schedule frames need  really strictly non-overlapping time periods (to obtain unambiguous schedules). In my opinion, the case of vehicle time frames is not similar.

In this case, the minute resolution is a bit vague to deal with. So it is good to verify it from the experts.
___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 164 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-Nij-UXs1QrCsU49mDNx:-Nio1jxAD05JbPg6LQz5:bdfp2sr) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L164)):*

Yep. The main point here is checking the integrity, that is, that each row (block or journey) does not overlap with the next one. For the last one there is no next one to compare against, there's no next row whose start it could overlap with. Previous row already checks overlap issues with the start of this last row that is discarded. So all is good.

Of course we could change things around a bit, do a `LEFT JOIN` instead, but that would just result in `block_overlaps_with_next` and `journey_overlaps_with_next` evaluating to `false` anyway, so no reason to complicate the logic because of that.

But maybe I'll add a comment that this has been thought of.
</blockquote></details>

Yeah. Then I read it correctly. A comment wouldn't hurt, because a careful reader will stop to think about it anyway.

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212)* status: 12 of 20 files reviewed, 14 unresolved discussions (waiting on @culka and @jarkkoka)

___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 37 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-NiiIbHw8sDoxLP5z8dv:-Nio8vZrCeyTMOF-knrH:b-6zqbkj) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L37)):*
> Now, you could get the arrival time of a stop point that is not the last in the journey.
I don't see how that would happen.

We have specified passing times so that first stop has `arrival_time = null AND departure_time <> null` and last stop has `departure_time = null AND arrival_time <> null` (not exactly with this code). Also the order in passing times and stop sequences must match, so we can't have out of order departure/arrival times. All of this is actually validated and tested as well, not just implicitly specified.

So first stop always has the smallest `departure_time` and last stop the largest `arrival_time`, which means that using just `MIN(tpt.departure_time)` and `MAX(tpt.arrival_time)` is enough, no coalescing with passing time needed. Or then I'm missing something.
___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 57 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-NiiU4fu1J2l-MxQK1te:-NioB56_Ateo8I83_Aub:b-tp44w) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L57)):*

What/where is the specification for determining journey start and end time?

To me it would be more intuitive to use `turnaround_time` in determining the journey start time. In Hastus import files, the field is called `trp_lay_min_strt`. In Finnish, it is "Terminaaliaika".

Respectively, t would be more intuitive to me to use `layover_time` in determining the journey end time. In Hastus import files, the field is called `trp_lay_min_end`. In Finnish, it is "Elpymisaika".

Are you sure about the code is correct here?
</blockquote></details>

The specification based on which I implemented this was wrong: it had layover and turnaround times flipped. 
This has now been corrected. Flipped the fields accordingly here as well.
___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 112 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-NiiyfSSDugnP7_sZVNX:-NioCM5s1GR3jBuwpWqe:b9emy1t) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L112)):*

It seems to that `SELECT DISTINCT` could be used here to return unique set of identifiers.
</blockquote></details>

Think so, added :ok_hand: 
___
*[`test/hasura/generic/timetablesdb/datasets/types.ts` line 102 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-Nij1gdg1TMVZ5QXtQaT:-NioCc9Y3LTyBrR89kYx:bff0rtr) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/test/hasura/generic/timetablesdb/datasets/types.ts#L102)):*

I would order `turnaround_time` before `layover_time` for clarity.
</blockquote></details>

Reordered all of these to match the actual order.
___
*[`test/hasura/generic/timetablesdb/integration-tests/data-integrity/vehicle-service/sequential-integrity.spec.ts` line 81 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-Nij2Qpb5z9-tluIBPHY:-NioEN8d4MgadRdhOgUj:b29rg5u) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/test/hasura/generic/timetablesdb/integration-tests/data-integrity/vehicle-service/sequential-integrity.spec.ts#L81)):*

I would change the order of these two properties for the sake of logical consistency and clarity (for all occurrences in this file).
</blockquote></details>

All instances of layover / turnaround were flipped in this file. Fixed.

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212)* status: 12 of 20 files reviewed, 14 unresolved discussions (waiting on @culka and @jarkkoka)

___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 155 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-NiizhlE8aLpPc--31ym:-NisUT9D4YPg3SycccM2:b3pbkps) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L155)):*

This is a bit complicated when the turnaround time and layover time are zeros. Vehicle schedule frames need  really strictly non-overlapping time periods (to obtain unambiguous schedules). In my opinion, the case of vehicle time frames is not similar.

In this case, the minute resolution is a bit vague to deal with. So it is good to verify it from the experts.
</blockquote></details>

By the way, within a single journey similar situation is allowed, that there is 0 minutes between stops. See https://github.com/HSLdevcom/jore4-hasura/blob/1419d8ecc3fe981c8170521fe84b5e21e0cd69bd/test/hasura/generic/timetablesdb/integration-tests/data-integrity/vehicle-journey/passing-times.spec.ts#L374

jarkkoka (changes requested):
> Reviewed 7 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212)* status: 19 of 20 files reviewed, 3 unresolved discussions (waiting on @culka and @Leitsi)

___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 37 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-NiiIbHw8sDoxLP5z8dv:-NisclvV1sqU2NkAsGFZ:b-yivzdt) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L37)):*

> Now, you could get the arrival time of a stop point that is not the last in the journey.
I don't see how that would happen.

We have specified passing times so that first stop has `arrival_time = null AND departure_time <> null` and last stop has `departure_time = null AND arrival_time <> null` (not exactly with this code). Also the order in passing times and stop sequences must match, so we can't have out of order departure/arrival times. All of this is actually validated and tested as well, not just implicitly specified.

So first stop always has the smallest `departure_time` and last stop the largest `arrival_time`, which means that using just `MIN(tpt.departure_time)` and `MAX(tpt.arrival_time)` is enough, no coalescing with passing time needed. Or then I'm missing something.
</blockquote></details>

Ok. You seem to be right that we are relying on such a constraint in the `jore4-hastus` code:

https://github.com/HSLdevcom/jore4-hastus/blob/96a8d0ccd3d1e5a8781c3ba6586629d6938bf371/src/main/kotlin/fi/hsl/jore4/hastus/service/importing/ConversionsFromHastus.kt#L244

However, this is not required in the database. Since we have chosen the approach of producing a generic timetables database for the outside world (at the client's request), I can see that adding a coalesce operation here would serve those possible use cases (outside the Jore4 domain) where the `jore4-hastus` service is not in use.

Leitsi (commented):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212)* status: 19 of 20 files reviewed, 3 unresolved discussions (waiting on @culka and @jarkkoka)

___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 37 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-NiiIbHw8sDoxLP5z8dv:-NishEc-1v5DXkpF4o6w:bqjz6s6) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L37)):*
> this is not required in the database

It is though, and on generic side. See https://github.com/HSLdevcom/jore4-hasura/blob/611d0a39235501b5706ab63e2d5a4a98ded0b266/migrations/generic/timetables/2000000000003_R_after_migrate_create_passing_times/up.sql#L127C2-L143C12

The comments on these fields also refer to this (though they could be a bit more specific, null is only allowed in this exact case):

> **arrival_time** IS 'The time when the vehicle arrives to the SCHEDULED STOP POINT. Measured as interval counted from the midnight of the OPERATING DAY. **When NULL**, only the departure time is defined for the passing time. E.g. in case **this is the first SCHEDULED STOP POINT of the journey.**';
> **departure_time** IS 'The time when the vehicle departs from the SCHEDULED STOP POINT. Measured as interval counted from the midnight of the OPERATING DAY. **When NULL**, only the arrival time is defined for the passing time. E.g. in case **this is the last SCHEDULED STOP POINT of the journey.**';

jarkkoka (approved):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212)* status: 19 of 20 files reviewed, 3 unresolved discussions (waiting on @culka and @Leitsi)

___
*[`migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql` line 37 at r2](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212#-NiiIbHw8sDoxLP5z8dv:-NisjK369H2JmGc2EDhf:bhe6b8r) ([raw file](https://github.com/HSLdevcom/jore4-hasura/blob/fef8e6742514c354dd1e2d1bb374fedce7af3b84/migrations/generic/timetables/2000000000008_R_after_migrate_create_vehicle_service_sequential_integrity_validation/up.sql#L37)):*

> this is not required in the database

It is though, and on generic side. See https://github.com/HSLdevcom/jore4-hasura/blob/611d0a39235501b5706ab63e2d5a4a98ded0b266/migrations/generic/timetables/2000000000003_R_after_migrate_create_passing_times/up.sql#L127C2-L143C12

The comments on these fields also refer to this (though they could be a bit more specific, null is only allowed in this exact case):

> **arrival_time** IS 'The time when the vehicle arrives to the SCHEDULED STOP POINT. Measured as interval counted from the midnight of the OPERATING DAY. **When NULL**, only the departure time is defined for the passing time. E.g. in case **this is the first SCHEDULED STOP POINT of the journey.**';
> **departure_time** IS 'The time when the vehicle departs from the SCHEDULED STOP POINT. Measured as interval counted from the midnight of the OPERATING DAY. **When NULL**, only the arrival time is defined for the passing time. E.g. in case **this is the last SCHEDULED STOP POINT of the journey.**';
</blockquote></details>

Ok, thanks for the clarification. Yes, it is there!

Had an IDE issue looking for it. Now resolved.

jarkkoka (approved):
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/212)* status: 19 of 20 files reviewed, 2 unresolved discussions (waiting on @culka and @Leitsi)

PR #220, opened 2024-01-05
Add hardcoded ids to vehicle types populated in migrations

Leitsi:
These missing meant that vehicle types got a random UUID each time. This makes creating test data a bit more difficult sometimes, because for blocks the vehicle_type_id is a required value. One example of this is the test-db-manager seed script, which has to fetch vehicle types from a running DB before it can create the seed data. If we hard code the ids, we can just use known values in such test code.

We already do this kind of hard coded UUIDs for values populated in migrations at least for service_calendar.day_type and service_calendar.day_type_active_on_day_of_week tables.

This only affects new databases and not eg. our test (demo) environment, but that is fine, since this fix is meant to address issues in test data and the id values are not relied upon in any app code.

There didn't seem to be any other tables that had a similar issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/220)


culka (approved):
:lgtm: 

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/HSLdevcom/jore4-hasura/220#-:-NnOxQ-M0P2tywsqNJHT:bg7igkv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Leitsi)

